PR #7, opened 2021-10-13
Remove broken and/or redundant logic from the `ModelInfo` object

Wuestengecko:
The properties of the `ModelInfo` objects (available e.g. as `MelodyModel.info`) are broken:

* `derive_version_url()` assumes that `url` points to a project on a Gitlab instance. We cannot simply make that assumption, and it breaks in several supported cases:
  - Remote repos stored anywhere that's not a Gitlab instance will result in incorrect URLs.
  - Local `file://` repositories or ones accessed over `ssh://` would result in a path that does not exist (however, see the next point as well).
* `url` is set to `None` when it's not an `https://` URL. This also breaks several different use cases:
  - It prevents the `LocalFileHandler` from storing the `path` in that field, as it will never be an `https://` URL.
  - Other URLs (e.g. `git+ssh://`, `git+file://`) are not saved.
* `as_dict()` is simply a call to `dataclasses.asdict(self)`. This looks very redundant to me, and I don't see a use case for that method to begin with (considering that `dataclass` does not provide that by default, and instead refers to the module-level function).

Historic (non-published) blame brings up two relevant commits, `fed7d084628f0216b607f1760185cc8bc2d83e35` by @MoritzWeber0 (introducing the first two issues) and `09b8dc46360f24eb531d93516259edaa8d0a8121` by @ewuerger (introducing the third). Neither commit provides any explanation on why the methods were introduced in the first place (only that they were, but well, that's a start I guess), nor why they have their specific logic.

Therefore, with this PR I propose to remove the relevant methods altogether. If a future `FileHandler` subclass finds itself in a position where it can – without unreasonable assumptions – provide similar additional information, it should make a subclass of `ModelInfo` and return that instead of adding its specific logic onto a very generic class.

### DIFF
--- a/capellambse/loader/modelinfo.py
+++ b/capellambse/loader/modelinfo.py
@@ -24,33 +24,3 @@ class ModelInfo:
     revision: str | None = None
     rev_hash: str | None = None
     capella_version: str | None = None
-
-    def __post_init__(self) -> None:
-        self.set_project_url()
-        self.version_url = self.derive_version_url()
-
-    def derive_version_url(self) -> str | None:
-        if self.url and self.rev_hash:
-            return "/".join(
-                (
-                    self.url,
-                    "-",
-                    "commit",
-                    self.rev_hash,
-                )
-            )
-        return None
-
-    def set_project_url(self) -> None:
-        """Set self.url to repository URL if the url is an HTTPS URL.
-        Otherweise set it to None.
-        """
-        if self.url:
-            if self.url.endswith(".git"):
-                self.url = self.url[: -len(".git")]
-
-            if not self.url.startswith("https://"):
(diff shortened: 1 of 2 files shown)


MoritzWeber0:
After reviewing the changes again, I can agree that the best solution would be moving it out of capellambse. 
Because of these big limitations, I would rather introduce it in the `TemplateHelper` section of `mddocgen` with a detailed mention of the limitations in the function definition (something like `derive_project_url_for_gitlab`). As the logic are of the functions is pretty simple and doesn't require any external tools like access to the `git` cli, we can just use the existing ModelInfo attributes.

MoritzWeber0 (approved):

PR #12, opened 2021-11-01
Requirements

amolenaar:
Updates for working with requirements


Wuestengecko:
@amolenaar When pulling from upstream just now, I noticed that you branched off of a fairly old `master` (~ 11 days before your first branch commit). Please take care to pull the latest upstream before starting to work in the future. This avoids all kinds of possible issues and mishaps outright, and makes for a cleaner [commit graph](https://github.com/DSD-DBS/py-capellambse/network). :)

PR #14, opened 2021-11-05
Fix issue when updating a property with the default value

amolenaar:
This failed if the value already was the default.

### DIFF
--- a/capellambse/extensions/reqif.py
+++ b/capellambse/extensions/reqif.py
@@ -427,7 +427,8 @@ class ValueAccessor(c.Accessor):
             raise TypeError("Value needs to be of type " + error_msg)
 
         if value == default:
-            del obj._element.attrib["value"]
+            if "value" in obj._element.attrib:
+                del obj._element.attrib["value"]
             return
 
         if isinstance(value, bool):


capellambse/model/common/element.py
```diff
@@ -192,6 +192,9 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
         return self._element is other._element
 
+    def __hash__(self):
+        return hash(self._element)
+
```
Wuestengecko, line 197: We had a discussion about this recently here: https://github.com/DSD-DBS/py-capellambse/pull/3#discussion_r719318211

In the comments to that PR, we hadn't actually reached a conclusion one way or the other (making it hashable vs. keeping it non-hashable). To summarize, here were the main arguments:

Pro: You can use it in hash sets.
Contra: It doesn't make semantic sense, since these objects are mutable (as in: they appear to be mutable to the end user, i.e. their attribute values can change during their lifetime). Conventionally, this means they should be non-hashable.

We had subsequently removed some set logic and changed it to use simple `list`s instead, which made me believe that the matter had been settled on "non-hashable" (see 8a27384a04ebcbcdebaf994a6b748a024fd2d9b1 and its follow-up b1a92d677b75a82fba093bb2b35369c5e529b099, as well as edfa673e6c486ccf4972ebd7cc2a2f15bd0a1a58).

Why this sudden change of heart?

And also, what does this have to do with the PR's topic ("Fix issue when updating a property with the default value")?

Wuestengecko (changes requested):

capellambse/model/common/element.py
amolenaar: As long as we define `__hash__` in terms of the immutable parts of the object, there's no problem.

Currently I have to work around the inability to use sets:

```python
{rcl.uuid: rlc for pc in model.pa.all_components for rlc in pc.realized_logical_components}.values()
```

Where it would be more natural to write:

```python
{rlc for pc in model.pa.all_components for rlc in pc.realized_logical_components}
```
> Why this sudden change of heart?

At that time I was looking at the code itself. Now that I use it actively, it stroke me as odd that I can not create sets.

> And also, what does this have to do with the PR's topic ("Fix issue when updating a property with the default value")?

Indeed, the PR name should be updated.

Wuestengecko: > it would be more natural to write:
> 
>     {rlc for pc in model.pa.all_components for rlc in pc.realized_logical_components}

The Google style guide has an opinion about that as well. :)

> Multiple `for` clauses or filter expressions are not permitted.
>
> — <https://google.github.io/styleguide/pyguide.html#27-comprehensions--generator-expressions>

Obviously this doesn't apply to code outside of this repo though. And as I mentioned in the other thread already, using a set was obviously the go-to first choice there as well.

And I agree with this choice. But we definitely need to make sure we actually use the new hashability in the places I mentioned in my previous comment. Can you take care of that? It makes sense if these follow-up changes happen quickly after the original change (and in the same PR).

amolenaar: I'll create a separate PR for element hashing.

Wuestengecko (approved):

PR #16, opened 2021-11-08
Fix test cases for Windows

amolenaar:
Taking one for the team 😉 .

Looks like two empty requirements are skipped now.

The tests that fail on windows are `XFAIL`'ed on other platforms. Something with rendering.


Wuestengecko (approved):
Thanks! This was long overdue.

> The tests that fail on windows are `XFAIL`'ed on other platforms.

Yes, because there are always some (sub-pixel) differences due to the exact choice of font family and library. It might be possible to check against that on a smaller scale like in the `svg` module tests (where we use ranges), but in a huge JSON document such as here it's a lot more to deal with.

PR #21, opened 2021-11-11
Make GenericElement hashable

amolenaar:
Having GenericElement as a hashable type (like `object`) allows us to work with sets of elements. We are not limited to using only lists.


Wuestengecko (approved):
Approved (as per previous discussions) :)

PR #22, opened 2021-11-11
Run Aird parser tests on all platforms

amolenaar:
The aird parser tests currently only run on Windows.

This PR adds the ability to test the parser and serializer on linux (working on macOS).

Although we should re-evaluate the value of those tests, having them running on all platforms is an improvement.

- [x] Tests run on Linux
- [x] Tests run on macOS


tests/test_model_loading.py
```diff
@@ -34,15 +40,9 @@ def test_model_loading_via_LocalFileHandler(path: str | pathlib.Path):
...
 
-@pytest.mark.parametrize(
-    "path",
-    [
-        pytest.param(
-            "git+" + pathlib.Path.cwd().as_uri(), id="From local repo"
-        ),
-    ],
-)
+@pytest.mark.skipif(not has_git_lfs(), reason="This test requires git-lfs to be present on the system")
```
Wuestengecko, line 43: I'd consider the hard requirement on Git LFS a bug. Although it has become quite common, we still can't just require everyone to install LFS even if they don't use it.

After a brief look at the affected code, it should be fairly straight-forward to catch the produced error and pretend that `git lfs ls-files` had simply listed nothing. Getting test coverage on both cases in a single test run might become annoying though :)

amolenaar: I created a follow-up issue #32.

This one "issue" should not hold back the other changes in this PR.

amolenaar:
Partly merged: aird tests are considered broken, so this commit has been left out.

Issue #29, opened 2021-11-23
Missing link from abstract function to state (both directions)


vik378:
Should work as shown in the picture below:
![image](https://user-images.githubusercontent.com/8280844/143058324-eb415924-0be4-4529-8184-fb046304a19a.png)

PR #31, opened 2021-11-24
Fix function state rels #29

vik378:
this implements a solution for #29


jamilraichouni (approved):
Ran the passing tests. Discussed the mod with the author.

PR #43, opened 2021-12-13
Make RequirementType and EnumValue hashable

amolenaar:
Hashing only works if the `__eq__` and `__hash__` methods are defined in the same class.

### DIFF
--- a/capellambse/extensions/reqif.py
+++ b/capellambse/extensions/reqif.py
@@ -464,6 +464,9 @@ class EnumValue(ReqIFElement):
             return self.long_name == other
         return super().__eq__(other)
 
+    def __hash__(self):
+        return super().__hash__()
+
 
 @c.xtype_handler(None, XT_REQ_TYPE_ENUM)
 class EnumDataTypeDefinition(ReqIFElement):
--- a/capellambse/extensions/reqif.py
+++ b/capellambse/extensions/reqif.py
@@ -523,6 +523,9 @@ class AbstractType(ReqIFElement):
 
         return super().__eq__(other)
 
+    def __hash__(self):
+        return super().__hash__()
+
 
 @c.xtype_handler(None, XT_MODULE_TYPE)
 class ModuleType(AbstractType):
--- a/tests/test_reqif.py
+++ b/tests/test_reqif.py
@@ -831,3 +831,10 @@ class TestRequirementsFiltering:
         assert isinstance(req_type, reqif.RequirementType)
         assert isinstance(req_type, t.Hashable)
         assert hash(req_type)
+
+    def test_enum_value_is_hashable(self, model: capellambse.MelodyModel):
+        enum_value = model.by_uuid("efd6e108-3461-43c6-ad86-24168339ed3c")
+
+        assert isinstance(enum_value, reqif.EnumValue)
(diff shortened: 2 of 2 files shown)


vik378 (approved):
looks good to me, lets merge it

Wuestengecko:
Hi! I came to this PR by tracing back the origin of a pylint warning that flew into my face, and I have opinions about it. :)

There's a good reason why Python requires an explicit override of the `__hash__` method when overriding `__eq__`. Just keeping the parent class' `__hash__` in this case violates the one fundamental requirement on `__hash__`:

> The only required property is that objects which compare equal have the same hash value
>
> — [Python Documentation on `object.__hash__()`](https://docs.python.org/3/reference/datamodel.html#object.__hash__)

Since instances of these classes compare equal to the string value of their `.long_name`, their hash value must be equal to that of the `long_name`; in other words `hash(some_enum_value) == hash(some_enum_value.long_name)` must always be `True`. Otherwise they cannot be looked up in hash collections (i.e. dictionaries) by their long_name, which is the exact property that we were after with having them compare equal in the first place. Try this with our 5.0 test model:

```python
>>> dtd = model.by_uuid("637caf95-3229-4607-99a0-7d7b990bc97f")
>>> dtd.values
<CoupledElementList at 0x00007F2D8FB9BCB0 [<EnumValue 'enum_val1' (efd6e108-3461-43c6-ad86-24168339ed3c)>, <EnumValue 'enum_val2' (3c2390a4-ce9c-472c-9982-d0b825931978)>]>
>>> "enum_val1" in dtd.values
True
>>> "enum_val1" in set(dtd.values)
False
```

Because the `__eq__()` of these two classes delegates to `GenericElement` through `super()`, the hash value of `EnumValue`s must also be equal to the hash of `self._element` (see https://github.com/DSD-DBS/py-capellambse/blob/a501bdf2a77ea906f79cc79bae793283a805dff3/capellambse/model/common/element.py#L206..L207). Satisfying both conditions at once is not possible for obvious reasons, but it's safe to drop the super delegation.

However, it is **not** safe to make the `hash()` based on the `.long_name`. Remember: The only reason we can get away with `GenericElement`s being hashable at all is that their `_element` is considered immutable over their lifetime. Everything else, specifically everything that accesses the XML, is mutable and can therefore not be used in hashes.

Therefore the only possible course of action for this pull request (that I can come up with) is being rejected (or reverted, as it has already found its way into master).

There are basically two ways how we can address the unhashability of `EnumValue` and `AbstractType`:

1. We just keep these unhashable
2. We remove the equality to their `.long_name` attribute

Both cause inconveniences for different use cases.

In my opinion, we should go with option 1 and advise users who unconditionally require the hashability of all model objects (independent of their type) to explore alternative solutions. To facilitate this, we can "open up" the `_element` a little bit. Currently we treat it as a private attribute; we could "publish" it as an opaque, but hashable object. Then advanced users could use it to make hashes of every possible model object. In this case, to make the user experience around it more pleasant, we should also offer a function that takes such an object and converts it back into a "proper" model object, i.e. a `GenericElement` instance. This would most likely be a method on `MelodyModel`, analogous to `.search()` and `.by_uuid()`.

Issue #49, opened 2022-01-14
Missing all_physical_paths at PA level


vik378:
In PA we are missing the physical path object. We should fix that 

![image](https://user-images.githubusercontent.com/8280844/149586934-e9ff0809-ea7d-43a7-8052-66c575fb5cc3.png)

This is what we should do to close this issue:

- [ ] add `.all_physical_paths` accessor to `PhysicalArchitecture`
- [ ] introduce `PhysicalPath` class into `cs.py`
- [ ] add `.physical_paths` accessor to `PhysicalLink`
- [ ] add tests for new functionality
- [ ] update API documentation

vik378:
This is how Capella actual meta-model looks like for `PhysicalPath`:
![image](https://user-images.githubusercontent.com/8280844/149588179-be441dba-8cb9-4189-9395-dbfe0bfd78ea.png)

PR #51, opened 2022-01-15
Implemented functionality for #49

vik378:
* PhysicalLink now has accessors for physical paths and allocated exchanges
* introduced PhysicalPath
* PhysicalPath now has accessors for involved links and allocated exchanges
* PhysicalArchitecture now has accessors for all_physical_paths and all_component_Exchanges
* added tests and test items for all new features
* updated API model

### DIFF
--- a/capellambse/model/crosslayer/cs.py
+++ b/capellambse/model/crosslayer/cs.py
@@ -25,14 +25,15 @@ Composite Structure object-relations map (ontology):
 import operator
 
 from capellambse.loader import xmltools
-from capellambse.model.crosslayer.fa import ComponentExchange
 
 from .. import common as c
-from . import capellacommon, information
+from . import capellacommon, fa, information
 
 XT_DEPLOY_LINK = (
     "org.polarsys.capella.core.data.pa.deployment:PartDeploymentLink"
 )
+XT_PHYS_PATH = "org.polarsys.capella.core.data.cs:PhysicalPath"
+XT_PHYS_PATH_INV = "org.polarsys.capella.core.data.cs:PhysicalPathInvolvement"
 
 
 class Component(c.GenericElement):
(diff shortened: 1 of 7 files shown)


tests/test_xlayer_cs.py
```diff
@@ -0,0 +1,39 @@
...
+    link_0 = model.pa.all_physical_links.by_name("Eth Cable 2")
+    link_1 = model.pa.all_physical_links.by_name("Eth Cable 3")
+    assert link_0 == path.involved_links[0]
+    assert link_1 == path.involved_links[1]
+
+
+def test_physical_path_has_exchanges(model: MelodyModel):
+    exchange = model.pa.all_component_exchanges.by_name("C 6")
+    path = model.pa.all_physical_paths.by_name("card1 - card2 connection")
+    assert exchange in path.exchanges
```
Wuestengecko, line 27: If there's only supposed to be this one exchange, we'd probably want to assert that `path.exchanges == [exchange]` or similar. That way we also ensure that we don't mistakenly find unrelated exchanges.

Same thought applies to the other tests following below.

ewuerger (approved):
Just improved structure of added classes to be consistent with other layers. For me personally, I like to define the xtype-strings on top of the file such that other layers that need this xtype can easily import them iff there is no class for it. As for the xtype_handler it is equivalent.

capellambse/model/crosslayer/cs.py
```diff
@@ -102,40 +104,39 @@ class PhysicalLink(PhysicalPort):
...
+        xtypes=fa.XT_COMP_EX_ALLOC,
         aslist=c.ElementList,
         follow="targetElement",
     )
 
+    physical_paths: c.Accessor
+
 
-@c.xtype_handler("org.polarsys.capella.core.data.cs:PhysicalPath")
+@c.xtype_handler(XT_PHYS_PATH)
```
Wuestengecko, line 117: This must be `None`; the first argument to `xtype_handler` is the architectural layer that this object can appear on (where `None` means it can appear anywhere, suitable for cross-layer objects).

Wuestengecko (changes requested):

tests/test_xlayer_cs.py
vik378: for now it is only this one but as the test model evolves there might be more and so we'll have this test failing

PR #58, opened 2022-01-26
Extending the _repr_html_ appearance for ReqIF

### LINKED ISSUE #38: Improve requirement object appearance in display context
By default a requirement object is described as GenericElement, however since it is closer to ReqIF element it needs to be described differently - the following attributes should be visible:

- [ ] RegIF long name
- [ ] ReqIF identifier
- [ ] ReqIf text
- [ ] ReqIF attributes (if any)
- [ ] links (if any)

ewuerger:
Close #38
Added additional wrapper for reused _short_html_ method
Added _short_html_ to ReqIFElement to handle naming via long_name
Added special _short_html_ to AbstractRequirementsAttribute to handle
naming via AttributeDefinition


capellambse/model/common/element.py
```diff
@@ -617,13 +623,15 @@ def __html__(self) -> markupsafe.Markup:
...
-            )
+            if hasattr(i, "_short_html_"):
+                content = i._short_html_()
+            else:
+                content = (
+                    f"<strong>{escape(type(i).__name__)}</strong>"
+                    f" &quot;{escape(i.name)}&quot;"
+                    f" ({escape(i.uuid)})"
+                )
+            fragments.append(f"<li>{content}</li>")
```
Wuestengecko, line 634: It should be safe to drop the condition and `else` branch here entirely and just assume that everything has a `_short_html_` method, because `_short_html_` is introduced into the hierarchy by `GenericElement` and everything yielded by `iter(self)` will be a subclass of that.

```suggestion
            fragments.append(f"<li>{i._short_html_()}</li>")
```

capellambse/model/common/element.py
```diff
@@ -262,10 +262,16 @@ def __html__(self) -> markupsafe.Markup:
...
     def _short_html_(self) -> markupsafe.Markup:
         return markupsafe.Markup(
+            self._wrap_short_html_(
+                f" &quot;{markupsafe.Markup.escape(self.name)}&quot;"
+                f"{(': ' + str(self.value)) if hasattr(self, 'value') else ''}"
+            )
+        )
+
+    def _wrap_short_html_(self, content: str) -> str:
+        return (
```
Wuestengecko, line 272: It would make much more sense (both semantically and from a code sharing perspective) to directly return a `markupsafe.Markup` object here instead of a `str` which then needs to be wrapped in `Markup` by the caller, seeing as this is the only use case anyway. (And it would save on a whole bunch of useless `)` lines too :) )

Wuestengecko (changes requested):

examples/Introduction to Requirement access and management with ReqIF extension.ipynb
```diff
@@ -0,0 +1,283 @@
...
+       "<CoupledElementList at 0x00007F2417F11A60 [<Requirement 'Test Module/This is a folder./<p>Test requirement 1 really l o n g text that is&nbsp;way too long to display here as that</p>\\n\\n<p>&lt; &gt; &quot; &#39;</p>\\n\\n<ul>\\n\\t<li>This&nbsp;is a list</li>\\n\\t<li>an unordered one</li>\\n</ul>\\n\\n<ol>\\n\\t<li>Ordered list</li>\\n\\t<li>Ok</li>\\n</ol>\\n' (3c2d312c-37c9-41b5-8c32-67578fa52dc3)>, <Requirement 'Test Module/This is a folder./TypedReq2' (0a9a68b1-ba9a-4793-b2cf-4448f0b4b8cc)>, <Requirement 'Test Module/This is a folder./Subfolder/TestReq3' (79291c33-5147-4543-9398-9077d582576d)>, <Requirement 'Test Module/TypedReq1' (85d41db2-9e17-438b-95cf-49342452ddf3)>, <Requirement 'Test Module/New showcase req' (1322936c-df67-4a3f-a53f-ceb804aa9549)>, <Requirement 'Next Module/TestReq' (1092f69a-5f3a-4fe6-a8fd-b2dffde90650)>]>"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "new_req = model.oa.requirement_modules[0].requirements.create(\"Requirement\", name=\"New showcase req\")\n",
```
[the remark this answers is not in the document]
ewuerger: Yep.This is true. Great catch!

PR #62, opened 2022-02-10
Exchange trace pa

ewuerger:
This pull-request makes the following additions:

## PhysicalArchitectureLayer
Comparing with the LogicalLayer capabilities were also missing. FunctionalExchange lookup was named in alignment with the other layers:
- component_package
- capability_package
- all_capabilities
- all_function_exchanges

## fa.FunctionalExchange
- allocating_component_exchange
- owner

## fa.ComponentExchange
- allocating_physical_path
- allocating_physical_link
- owner

## cs.Component
- physical_paths
- physical_links

Please make sure to have a look at the `owner` property implementation. I tried to prepare write support by adding appropriate setters and deleters. I couldn't type annotate the owner attribute on the AbstractExchange class unfortunately. Mypy didn't let me.
The `owner.deleter` on the ComponentExchange is using the not yet implemented deleter of the `ReferenceSearchingAccessor`. What I really want here is to do the following:
```python
self.allocating_physical_path = self.allocating_physical_link = None
```
i.e. using the setter of the `ReferenceSearchingAccessor`... **if it is fine to set allocating_physical_path and allocating_physical_link to None when setting the owner**. One could also raise an `ValueError` and refer the user to set either `allocating_physical_path` or `allocating_physical_link`. What do you think @amolenaar, @Wuestengecko and @vik378?

edit: For the sceptics:

### DIFF
--- a/capellambse/model/crosslayer/fa.py
+++ b/capellambse/model/crosslayer/fa.py
@@ -159,18 +159,6 @@ class FunctionalExchange(AbstractExchange):
     def owner(self) -> ComponentExchange | None:
         return self.allocating_component_exchange
 
-    @owner.setter
-    def owner(self, exchange: ComponentExchange | None) -> None:
-        if not isinstance(exchange, (ComponentExchange, type(None))):
-            raise ValueError(
-                "Owner needs to be of type 'ComponentExchange' or 'NoneType'"
-            )
-        self.allocating_component_exchange = exchange
-
-    @owner.deleter
-    def owner(self) -> None:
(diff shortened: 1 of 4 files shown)


ewuerger:
The case where the PhysicalPath is preferred over the PhysicalLink is failing, due to my inability to alter the model such that there is a ComponentExhange that is allocated by a PhysicalLink and a PhysicalPath at the same time.

Please fix this @vik378.

vik378 (approved):
looks ok

PR #69, opened 2022-03-11
Added method _short_html_ to class AbstracDiagram

henrik997:
With it, diagrams and corresponding uuid can be listed.

### DIFF
--- a/capellambse/model/diagram.py
+++ b/capellambse/model/diagram.py
@@ -97,6 +97,9 @@ class AbstractDiagram(metaclass=abc.ABCMeta):
             + self.name
             + markupsafe.Markup("</figcaption></figure>")
         )
+    
+    def _short_html_(self) -> markupsafe.Markup:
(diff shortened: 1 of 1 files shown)


capellambse/model/diagram.py
```diff
@@ -97,6 +97,9 @@ def __html__(self) -> markupsafe.Markup:
             + self.name
             + markupsafe.Markup("</figcaption></figure>")
         )
+    
+    def _short_html_(self) -> markupsafe.Markup:
+        return markupsafe.Markup(f"<b>{self.name}</b> (uuid: {self.uuid if self.uuid else 'not applicable'})")
```
Wuestengecko, line 102: > `<b>{self.name}</b>`

`self.name` needs to be escaped here, in case it contains XML special characters. (See other `_short_html_()` or `__html__()` implementations for hints.)

Wuestengecko (changes requested):
Please make sure to run the `black` code formatter before committing (you should also look into setting up [`pre-commit`](https://pre-commit.com/)).

PR #70, opened 2022-03-16
Library notebook

henrik997:
Add a capella project using libraries and a corresponding notebook.


examples/05 Introduction to Libraries.ipynb
```diff
@@ -0,0 +1,382 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Product\n"
+     ]
+    }
+   ],
+   "source": [
+    "for component in model.oa.all_classes:\n",
+    "    print(component.name)"
```
Wuestengecko, line 353: This seems very redundant with lines 143f. What's the difference here, except that one addresses `model.la` and the other `model.oa`? If there is none, we can just drop this cell.

examples/05 Introduction to Libraries.ipynb
```diff
@@ -0,0 +1,382 @@
...
+     "text": [
+      "* <ExchangeItem 'ordered products' (b852ca69-cb56-40a2-a1ab-91c351ddabfe)>\n"
+     ]
+    }
+   ],
+   "source": [
+    "supply_details = model.la.all_function_exchanges[1]\n",
+    "print(supply_details.exchange_items)"
+   ]
+  },
```
Wuestengecko, line 292: These are standard model functions and have nothing to with libraries (AFAICT). If so, we should trim these cells away.

PR #71, opened 2022-03-23
Add Access History for ModelElements and Requirements

ewuerger:
# Access History for ModelElements

Whenever you are working with instances of ModelElements, for e.g. during template rendering of Jinja2, you might want to see which Elements were used during the process. The resulting ID container is restricted by a set of attribute names such that during audit the `uuid` is only added if the accessed attribute-name matches one of them.  

This PR implements an `AttributeAuditor` with following example:
```python
>>> with AttributeAuditor(model, {"name", "description"}) as auditor:
...     print(model.la.all_components[0].name)
...     print(model.la.all_components[1].uuid)
...
Hogwarts
6583b560-6d2f-4190-baa2-94eef179c8ea
>>> auditor.recorded_ids
{'0d2edb8f-fa34-4e73-89ec-fb9a63001440'}
```


capellambse/auditing.py
```diff
@@ -0,0 +1,78 @@
...
+        self.model = None
+
+    def __audit(self, event: str, args: tuple[t.Any, ...]) -> None:
+        if event == "capellambse.read_attribute":
+            obj, attr_name, _ = args
+            if not hasattr(obj, "_model") or obj._model is not self.model:
+                return
+
+            if attr_name in self.attrs and hasattr(obj, "uuid"):
+                self.recorded_ids.add(obj.uuid)
```
Wuestengecko, line 78: This will still cause a RecursionError when `uuid` is among the attributes to watch (either by explicitly specifying it, or by not specifying any attributes at all). We need to explicitly handle the `uuid` attribute if that is what is being accessed.

tests/test_auditing.py
```diff
@@ -0,0 +1,31 @@
...
+import capellambse
+from capellambse import auditing
+
+
+def test_AttributeAuditor_tracks_uuids_of_accessed_ModelElements(
+    model: capellambse.MelodyModel,
+):
+    # pylint: disable=pointless-statement
+    comp, comp1, comp2 = model.la.all_components[:3]
+    uuids = {comp.uuid, comp2.uuid}
```
Wuestengecko, line 25: A better name here IMO would be `expected_uuids` (or just `expected`, as is used in various other tests).

Also, for clarity, an empty line after this one would be good, to signify the end of the "setup phase" for this test.

Wuestengecko (changes requested):

ewuerger:
What do you think about the latest changes? @Wuestengecko 
Do you know how to fix it that mypy and Pylint in the workflows break?

capellambse/auditing.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Examples
+    --------
+    >>> auditor = AttributeAuditor(model, {"name", "description"})
+    >>> print(model.la.all_components[0].name)
+    Hogwarts
+    >>> auditor.recorded_ids
+    {'0d2edb8f-fa34-4e73-89ec-fb9a63001440'}
+    >>> # Cleanup
+    >>> auditor.model = None
```
Wuestengecko, line 43:
```suggestion
    >>> auditor.detach()
```

capellambse/auditing.py
```diff
@@ -0,0 +1,86 @@
...
+
+    >>> with AttributeAuditor(model, {"name", "description"}) as auditor:
+    ...     print(model.la.all_components[0].name)
+    ...
+    Hogwarts
+    >>> auditor.recorded_ids
+    {'0d2edb8f-fa34-4e73-89ec-fb9a63001440'}
```
Wuestengecko, line 50: What do you think about changing the interface of the context manager so that it directly gives the set of `recorded_ids` to the `as` keyword? The only reason that I can see why one would need the auditor object is to `detach()` it afterwards, which is done by the context manager automatically anyways. Directly handing over the set of IDs makes the usage a little more straight forward.

capellambse/auditing.py
```diff
@@ -0,0 +1,86 @@
...
+        self.model = None
+
+    def __audit(self, event: str, args: tuple[t.Any, ...]) -> None:
+        if event == "capellambse.read_attribute":
+            obj, attr_name, _ = args
+            last_obj, last_attr = self.last_audit
+            if obj is last_obj and attr_name == last_attr:
+                return
+
+            self.last_audit = (obj, attr_name)
```
Wuestengecko, line 81: Storing the `obj` here creates a transitive strong reference to the `model`. This must be cleaned up during `detach()`.

tests/test_auditing.py
```diff
@@ -0,0 +1,54 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+import capellambse
+from capellambse import auditing
```
Wuestengecko, line 17: I think it makes more sense for our tests to use the `AttributeAuditor` export on `capellambse`, as that is what's supposed to be the "public API" of it.

capellambse/auditing.py
```diff
@@ -0,0 +1,86 @@
...
+        self.recorded_ids: set[str] = set()
+        self.last_audit = (None, "")
+
+        sys.addaudithook(self.__audit)
+
+    def __enter__(self) -> AttributeAuditor:
+        return self
+
+    def __exit__(self, *_: t.Any) -> None:
+        return self.detach()
```
Wuestengecko, line 69: Redundant `return`.

```suggestion
        self.detach()
```

Wuestengecko (changes requested):

capellambse/auditing.py
```diff
@@ -59,18 +58,27 @@ def __init__(
...
+
+    def detach(self) -> None:
         self.model = None
 
     def __audit(self, event: str, args: tuple[t.Any, ...]) -> None:
         if event == "capellambse.read_attribute":
             obj, attr_name, _ = args
+            last_obj, last_attr = self.last_audit
+            if obj is last_obj and attr_name == last_attr:
```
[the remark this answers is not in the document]
ewuerger: Partially correct (like 90%)... We still run into recursion like this when we keep the `if hasattr(obj, "uuid")` in the last if condition. That is why we drop it and get much more simple like you advised (KISS - [Keep it simple, stupid](https://en.wikipedia.org/wiki/KISS_principle)).

Wuestengecko (approved):
Looks good. LGTM.

Issue #64, opened 2022-02-17
SAB diagram - multiple rendering issues


vik378:
SAB diagram rendered by our default engine has missing features and incorrect placement issues:

* Function ports are not shown
* ComponentExchange edges go into and above the Component Ports
* port allocation links go above ports
* Component Port direction is identified erroneously 

![image](https://user-images.githubusercontent.com/8280844/154484855-6a7028a7-4b7b-435e-9cbf-ec5f532621ab.png)

Wuestengecko:
Should be fixed with merge of #72. (If not, please reopen.)

PR #72, opened 2022-03-24
Fix edge port snapping

henrik997:
This PR fixes wrong placements of ports and ingoing edges for  `aird.Box` (relates to issue #64). Methods of `aird.Box` are adjusted such that edges can have three different routing styles: _oblique_, _rectilinear_ or _tree_. In case no bendpoints are available for an edge, they are generated by rerouting according to its routing style.

The diagram element `aird.Edge` has now the property decorators `center` and `length` where the first one describes the center of the middlemost segment of this edge and the second the total length.


capellambse/aird/diagram.py
```diff
@@ -620,6 +699,28 @@ def points(self) -> aird.Vec2List:
...
+        if len(self.points) == 2:
+            return (self.points[0] + self.points[1]) / 2
+
+        distance: float = 0.0
+        half_length: float = self.length / 2
+        for i in range(1, len(self.points)):
+            distance += (self.points[i - 1] - self.points[i]).length
+            if distance > half_length:
+                return (self.points[i - 1] + self.points[i]) / 2
+        return (self.points[-2] + self.points[-1]) / 2
```
Wuestengecko, line 722: While documented, the behavior of this property is all but intuitive to me. When I ask for an edge's `center`, I expect to receive the point that's exactly in the middle of the edge and not some point that might be arbitrarily far away from it. Consider the following very simple example:

```
>>> from capellambse.aird import Edge
>>> e = Edge([(0, 0), (100, 0), (100, 100)])
>>> print(e)
Edge between (0, 0) and (100, 100) via (100, 0)
>>> e.center
Vector2D(x=100.0, y=50.0)
```

The edge goes 100 pixels to the right and then 100 pixels down. It should be rather obvious that the edge's actual center should be exactly at the corner, i.e. `(100, 0)`. However the "center" that was calculated here is 50 pixels away from that.

While I can clearly see a use case for the actual center point of an edge (we even use it in the AIRD parser), I fail to see a use case here. Do you mind explaining your thoughts a bit?

Wuestengecko (changes requested):

capellambse/aird/diagram.py
henrik997: The name might be a little bit misleading, but you receive the center of the middlemost segment as explained in the docstring. During `route_manhattan` arguments `target` and `source` could potentially be of type `aird.Edge` and then, they would need the property `center`.

Wuestengecko: > they would need the property `center`.

They need it to *exist*, but it doesn't seem to me that they have strict requirements on what this property actually returns (other than "it must be a `Vector2D`"). So I see nothing speaking against doing "the right thing" (as in, what a non-doc-reading user would expect here). In addition to that, we basically have a working algorithm for this in the `parser` module already anyways, and it shouldn't take too much brain power to adapt.

PR #77, opened 2022-04-06
Code Generation

henrik997:
This PR gives an introduction how to generate class interfaces from _class diagrams_. For this, three different types of interfaces are provided: a _ROS message_, a _python class_ and a _protobuf_ interface.


examples/06 Code Generation.ipynb
```diff
@@ -0,0 +1,430 @@
...
+    "\n",
+    "    for counter, prop in enumerate(cls.properties):\n",
+    "        multiplicity = (\"\", \"[]\")[prop.max_card.value > 1.0]\n",
+    "        if (\n",
+    "            isinstance(\n",
+    "                prop.type, capellambse.model.crosslayer.information.Class\n",
+    "            )\n",
+    "            and prop.type not in current_classes\n",
+    "        ):\n",
+    "            current_classes.append([prop.type])\n",
```
Wuestengecko, line 345: You're checking whether `prop.type` is in current_classes, and if it's not, you're adding `[prop.type]` (note the brackets). Same thing where `[cls]` is added to the list earlier on. This means the containment check will still fail the next time we arrive here with the same argument, as it's not `prop.type` that's in `current_classes`, but rather a one-element list containing it. Try the following:

```python
>>> proptype = object()
>>> classes = []
>>> classes.append([proptype])
>>> proptype in classes
False
```

Wuestengecko (changes requested):

examples/07 Code Generation.ipynb
```diff
@@ -295,99 +285,90 @@
...
     "            )\n",
     "        else:\n",
-    "            text.append(\n",
-    "                f\"{indent}{prop.type.name}{multiplicity} {prop.name} = {counter+1};\"\n",
+    "            lines.append(\n",
+    "                f\"{indent}{prop.type.name}{multiplicity} {prop.name} = {counter};\"\n",
     "            )\n",
-    "    text[-1] = text[-1] + f\"\\n{indent[:-1]}}}\"\n",
-    "    text = \"\\n\".join(text)\n",
+    "    lines[-1] = lines[-1] + f\"\\n{indent[:-1]}}}\"\n",
```
Wuestengecko, line 338: You can use `+=` here. But really, you should rather use `lines.append()` instead.

Wuestengecko, line 338: Previous comment about `indent[:-1]` applies here as well. Looks like another left-over.

examples/07 Code Generation.ipynb
```diff
@@ -295,99 +285,90 @@
...
+      "        float lat = 1;\n",
+      "        float lon = 2;\n",
+      "        float alt = 3;\n",
+      "           message  Example {\n",
+      "            str test = 1;\n",
+      "           }\n",
+      "        repeated Example[] examples = 4;\n",
+      "       }\n",
+      "    repeated Waypoint[] waypoints = 1;\n",
+      "   }\n",
```
Wuestengecko, line 371: For some reason, everything is indent with 3 spaces now (and 4 additional spaces per sub-indent). That looks very odd, and might be a bug from a partial refactor. What do you say?

Wuestengecko (changes requested):
That looks a lot better already! :) Only some minor issues left (some of which were already present, but I hadn't noticed them last time).

examples/07 Code Generation.ipynb
henrik997: Okay, I use `+=`, because `lines[-1] is a `str` object which has no attribute `append` :)

Wuestengecko (approved):

PR #84, opened 2022-05-12
Exchange items filter

ewuerger:
This PR fixes the `ExchangeItem` filter and adds additonal control to handle sorting of rendered exchange items in the label.
The latter is inspired by the rcParams way of matplotlib... but they also have corresponding setters that essentially update the rcParams dictionary for plot attributes. 

In an earlier stage there were tests that used `get_model_element_from_diag_element`. This is a function that finds the model element from an aird diagram element by finding the target element underneath and use the href attribute. The other way around  does not work uniquely, since there can be multiple target elements with the same referenced UUID under different diagrams. 
Open question to delete this function or leave it in the helpers. It's not used for now.

To test that the filters work we need diagrams where the filters were applied. I could have just copied the diagram in capella and applied the individual filters. This way would inflate the test model with unnecessary duplications of diagrams... Not my preferred style. That is why we introduce [`activated_filters`](https://github.com/DSD-DBS/py-capellambse/commit/77dbc260aa48942dfc69ff2bec2a9db9529b4f63#diff-907234b3a22dbde03b41bc0d89488a36fca1f072fcc1a0652576203e11eff9ccR305-R334) with corresponding add and remove functions that insert/remove `<activeFilters>` XML elements before the `<filtersHistory>` XML element. These elements store the relevant filter name in their href attribute (`@filters="show.exchange.items"`). The `activated_filters` attribute returns a list of names of all active filters on the diagram. 

This new feature makes it very convenient to apply filters headlessly. The only downside is, that this works for only supported filters which are registered in `capellambse.aird.parser._filters.GLOBAL_FILTERS`. Make sure to add/remove filters **before** rendering diagrams such that the `capellambse.aird.parser._filters.applyfilters` is called.

Example:

```python
diag = model.diagrams.by_name("My diagram")
diag.add_activated_filter("show.exchange.items.filter")  # resulting label will only show ExchangeItems -> [E1, E2,...]
diag = diag.render(None)
```


capellambse/model/diagram.py
```diff
@@ -302,6 +302,36 @@ def type(self) -> modeltypes.DiagramType:
...
+        Writes a new <activatedFilters> XML element to the
+        <diagram:DSemanticDiagram> XML element. If the `fltr_name` is
+        not apparent in :data:`aird.parser.GLOBAL_FILTERS` as a key
+        it can not be applied when rendering. It should still be visible
+        in the GUI.
+        """
+        aird.parser._filters.add_filter_to_diagram(
+            self._model._loader, self._element, fltr_name
+        )
+        self._create_diagram()
```
Wuestengecko, line 324: Note that `self._create_diagram()` *returns* the rendered diagram, however the return value is not used in this method. Therefore this call is superfluous. I assume what you wanted to do here was to refresh the render cache.

Instead, we should merely *invalidate* it so that a full re-render will happen on the next call to `render()`. Otherwise multiple calls to `add_activated_filter` would cause multiple intermediate renders to happen (one for each added filter), which is a significant waste of resources. And on top of that, there's the possible use case of programmatically adding the filters on several diagrams at once and not rendering them with capellambse at all, but saving the filters for further use in Capella. (Adding a public method on `AbstractDiagram` to `invalidate_cache()` seems like an overdue change anyways, so let's do that while we're at it.)

I believe this should also address your comment from the opening post:

> Make sure to add/remove filters **before** rendering diagrams such that the `capellambse.aird.parser._filters.applyfilters` is called.

Because, after invalidating the cache, this function would in fact be called again during the next `render()`.

Wuestengecko (changes requested):
To also address some more general points from the opening post as well:

> In an earlier stage there were tests that used `get_model_element_from_diag_element`. […] It's not used for now.

+1 for removing it. Maintaining random functions that we don't use creates unnecessary overhead.

> To test that the filters work we need diagrams where the filters were applied. […] That is why we introduce [`activated_filters`](https://github.com/DSD-DBS/py-capellambse/commit/77dbc260aa48942dfc69ff2bec2a9db9529b4f63#diff-907234b3a22dbde03b41bc0d89488a36fca1f072fcc1a0652576203e11eff9ccR305-R334) with corresponding add and remove functions that insert/remove `<activeFilters>` XML elements before the `<filtersHistory>` XML element.

Good choice indeed. These functions are small and focused enough to not cause bigger problems, and they're the first step towards actually modifying AIRD data (in case we ever want to fully get there).

Did you test whether it's actually necessary for the `<activeFilters>` to appear before the `<filtersHistory>`? It might just as well be that this ordering is an implementation detail somewhere deep in Capella, and they would actually work fine if we were to just append them at the end. This would avoid the need to search for some specific place to insert them, and make our code a little more robust.

> This new feature makes it very convenient to apply filters headlessly. The only downside is, that this works for only supported filters which are registered in `capellambse.aird.parser._filters.GLOBAL_FILTERS`.

As the filter names in the XML are quite cryptic and arbitrary at times, it might be a good idea to add an `Enum` or something that has all the filters that exist. Each member could have a short, but ideally still readable name, and map to the internal XML name. Then each docstring could have the full name shown in the GUI.

capellambse/aird/parser/_filters/__init__.py
```diff
@@ -265,6 +269,62 @@ def _extract_filter_type(flt_elm: etree._Element) -> str:
     return urllib.parse.unquote(flttype.group(1))
 
 
+def get_filters_from_diagram(
+    melodyloader: capellambse.loader.MelodyLoader, uuid: str
+) -> list[str]:
+    diag_elt = melodyloader[uuid]
```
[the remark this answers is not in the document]
ewuerger: Very good point. Have a look on the new `ActiveFilters` set-object that you inspired me to make. This is using the `DiagramDescriptor.uid` to receive the diagram element via the loader. We can see how this looks after the requested changes. =)

capellambse/aird/parser/_filters/__init__.py
```diff
@@ -265,6 +269,62 @@ def _extract_filter_type(flt_elm: etree._Element) -> str:
...
+
+def add_filter_to_diagram(
+    melodyloader: capellambse.loader.MelodyLoader,
+    diag_descriptor: DiagramDescriptor,
+    filter_name: str,
+) -> None:
+    try:
+        GLOBAL_FILTERS[filter_name]
+    except KeyError:
+        LOGGER.warning("This filter is not yet supported")
```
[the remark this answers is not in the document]
ewuerger: Removed this completely like suggested.

capellambse/model/crosslayer/__init__.py
```diff
@@ -6,7 +6,7 @@
 .. diagram:: [CDB] BaseLayer
 """
 from .. import common as c
-from . import cs, information
+from . import capellacommon, capellacore, cs, fa, information, interaction
```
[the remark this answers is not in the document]
ewuerger: This is for being able to import them from the crosslayer for type-annotations.

capellambse/model/diagram.py
ewuerger: We did it a bit different. We should have a talk/call about this.

capellambse/aird/parser/_filters/global.py
```diff
@@ -7,39 +7,55 @@
...
-        assert obj.uuid is not None
-        _, items = _get_allocated_exchangeitem_names(
-            obj.uuid,
-            alloc_attr="exchangedItems",
-            melodyloader=melodyloader,
+        assert isinstance(label.label, str)
+        label.label += " " + _stringify_exchange_items(obj, melodyloader)
+
+
+@global_filter("show.exchange.items.filter")
```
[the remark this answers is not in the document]
ewuerger: Resolved by aggreeing to pursue issue #86.

capellambse/aird/parser/_filters/__init__.py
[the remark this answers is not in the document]
ewuerger: To be resolved in opened issue #87.

capellambse/aird/parser/_filters/__init__.py
```diff
@@ -42,6 +41,8 @@
 GLOBAL_FILTERS: dict[str, GlobalFilter] = {}
 LOGGER = logging.getLogger(__name__)
 
+XP_PLUGIN = "/plugin/org.polarsys.capella.core.sirius.analysis/description/context.odesign#/"
```
Wuestengecko, line 44: The prefix `XP_` suggests (and is used in other places for) an `etree.XPath` object. This however is not an XPath, it's just a simple "magic" string that we need to insert in the right place. (Call it platform path if you will, but that's still not an XPath.)

capellambse/model/diagram.py
```diff
@@ -303,37 +318,42 @@ def type(self) -> modeltypes.DiagramType:
             return modeltypes.DiagramType.UNKNOWN
 
     @property
-    def activated_filters(self) -> list[str]:
-        """Return a sequence of activated filters"""
-        return aird.parser._filters.get_filters_from_diagram(
-            self._model._loader, self._element.uid
-        )
+    def filters(self) -> aird.parser._filters.ActiveFilters:
```
Wuestengecko, line 321: I'm not happy with the return annotation here. First, it can be simplified as `aird.ActiveFilters`, but IMO we can easily consider this entire class to be an implementation detail and just announce it as `cabc.MutableSet[str]`. And then we don't even need to expose the `ActiveFilters` class publicly in the first place.

capellambse/model/diagram.py
```diff
@@ -303,37 +318,42 @@ def type(self) -> modeltypes.DiagramType:
...
-        )
+        return self._render_params
 
     def _create_diagram(self) -> aird.Diagram:
-        return aird.parse_diagram(self._model._loader, self._element)
+        if self._element.render_params != self._render_params:
+            self._element.render_params = self._render_params
+
+        diagram = aird.parse_diagram(self._model._loader, self._element)
+        self._force_fresh_rendering = False
```
Wuestengecko, line 355: Left-over old cache invalidation.

tests/test_aird_filters.py
```diff
@@ -4,66 +4,86 @@
...
-    "ModelExtensionFilter",
-    "hide.overlappedfunctional.chains.label.filter",
-    "hide.overlappedfunctional.chains.icon.filter",
-]
+DEFAULT_ACTIVATED_FILTERS = frozenset(
+    [
+        "ModelExtensionFilter",
+        "hide.overlappedfunctional.chains.label.filter",
+        "hide.overlappedfunctional.chains.icon.filter",
+    ]
```
Wuestengecko, line 18:
```suggestion
    {
        "ModelExtensionFilter",
        "hide.overlappedfunctional.chains.label.filter",
        "hide.overlappedfunctional.chains.icon.filter",
    }
```

tests/test_aird_filters.py
```diff
@@ -2,18 +2,96 @@
...
+from capellambse import MelodyModel, aird
 
 COMP_PORT_FILTER_DIAG = "[LAB] Test Component Port Filter"
+EX_ITEMS_FILTER_DIAG = "[SDFB] Test ExchangeItem Filter"
+STD_ACTIVATED_FLTERS = [
+    "ModelExtensionFilter",
+    "hide.overlappedfunctional.chains.label.filter",
+    "hide.overlappedfunctional.chains.icon.filter",
+]
+EX_ITEMS_FLTER = "show.exchange.items.filter"
```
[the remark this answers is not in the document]
Wuestengecko: Closed without comment and without code changes? How is this resolved?

tests/test_aird_filters.py
```diff
@@ -4,66 +4,86 @@
...
+def test_remove_activated_filter_on_diagram(
+    model_5_2: capellambse.MelodyModel, filter_name: str
+) -> None:
+    diag: _m.diagram.Diagram = model_5_2.diagrams.by_name(EX_ITEMS_FILTER_DIAG)
 
+    diag.filters.remove(filter_name)
 
-def test_remove_activated_filter_on_diagram(model_5_2: MelodyModel) -> None:
-    diag = model_5_2.diagrams.by_name(EX_ITEMS_FILTER_DIAG)
+    assert filter_name not in diag.filters
```
Wuestengecko, line 50: This assertion can easily be made even stricter:

```suggestion
    assert diag.filters == DEFAULT_ACTIVATED_FILTERS - {filter_name}
```

tests/test_aird_filters.py
```diff
@@ -2,18 +2,96 @@
...
         if d.port and d.styleclass and d.styleclass.startswith("CP_")
     )
+
+
+def test_fex_exchangeitems_filter_is_applied(model_5_2: MelodyModel) -> None:
+    diag = model_5_2.diagrams.by_name(EX_ITEMS_FILTER_DIAG)
+    aird.RENDER_PARAMS["sorted_exchangedItems"] = False
+
+    diag.add_activated_filter(EX_ITEMS_FLTER)
+    diag = diag.render(None)
```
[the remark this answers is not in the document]
Wuestengecko: My main point was that you were re-using a variable with a different type, which is only "valid" (i.e. mypy doesn't complain about it) because both assignments result in `typing.Any` and therefore are not type-checked at all.

The explicit annotations (to resolve `Any` to a concrete type) are a nice sprinkle on top, and probably a good idea generally when `Any` is involved. Although `typing.Any` will cripple type checks in every case, so they're still only of limited use.

capellambse/model/crosslayer/__init__.py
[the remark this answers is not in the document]
Wuestengecko: I still don't get what the issue is. You can just say `from capellambse.model.crosslayer import capellacommon`, regardless of whether or not this import exists here, can you not? `cs` and `information` are necessary because they're used in code that lives in this specific file, but the new imports still look redundant to me.

Wuestengecko (changes requested):

capellambse/model/diagram.py
```diff
@@ -259,6 +272,20 @@ class Diagram(AbstractDiagram):
...
+    Additional rendering parameters.
+
+    Rendering options for :class:`aird.Diagram`s. Changing them will
+    force a fresh rendering of the diagram, neglecting the cache.
+
+    The following parameters are currently supported:
+
+        - ``sorted_exchangedItems`` (*bool*): Enable ExchangeItem
+        sorting when rendering diagrams with active ExchangeItems
+        filter (``show.exchange.items.filter``).
```
Wuestengecko, line 287:
```suggestion
        - ``sorted_exchangedItems`` (*bool*): Enable ExchangeItem
          sorting when rendering diagrams with active ExchangeItems
          filter (``show.exchange.items.filter``).
```

capellambse/model/diagram.py
```diff
@@ -302,8 +328,21 @@ def type(self) -> modeltypes.DiagramType:
...
+    @property
+    def filters(self) -> t.MutableSet[str]:
+        """Return a set of currently activated filters on this diagram."""
+        return aird.ActiveFilters(self._model, self)
+
+    @filters.setter
+    def filters(self, filters: cabc.Iterable[str]) -> None:
+        active_filters = aird.ActiveFilters(self._model, self)
+        active_filters.clear()
+        self.invalidate_cache()
```
Wuestengecko, line 340: Duplicate cache invalidation; the `active_filters.clear()` has already invalidated the cache before. (That's its job after all.)

```suggestion
```

capellambse/model/diagram.py
```diff
@@ -302,8 +328,21 @@ def type(self) -> modeltypes.DiagramType:
...
-    def _create_diagram(self) -> aird.Diagram:
-        return aird.parse_diagram(self._model._loader, self._element)
+    @property
+    def filters(self) -> t.MutableSet[str]:
+        """Return a set of currently activated filters on this diagram."""
+        return aird.ActiveFilters(self._model, self)
+
+    @filters.setter
+    def filters(self, filters: cabc.Iterable[str]) -> None:
+        active_filters = aird.ActiveFilters(self._model, self)
```
Wuestengecko, line 338: For simplicity, we can delegate to the property getter.

```suggestion
        active_filters = self.filters
```

tests/test_aird_filters.py
```diff
@@ -2,18 +2,129 @@
...
+def test_fex_exchangeitems_filter_is_applied(
+    model_5_2: capellambse.MelodyModel,
+) -> None:
+    diag: model_.diagram.Diagram = model_5_2.diagrams.by_name(
+        EX_ITEMS_FILTER_DIAG
+    )
+
+    diag.filters.add(EX_ITEMS_FILTER)
+    diagram = diag.render(None, sorted_exchangedItems=False)
+    fex_edge = diagram["_yovTvM-ZEeytxdoVf3xHjA"]
```
Wuestengecko, line 95: Considering the volatile nature of these auto-generated IDs, do you suppose it's a good idea to identify this object in a different way maybe? Perhaps make a simple diagram that only has this one single FEX displayed, and identify it by its type (edge) and styleclass (FunctionalExchange). That would prevent this test from breaking when Capella inevitably rewrites the IDs as part of normal usage.

tests/test_helpers.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 
-from capellambse import helpers
+from capellambse import MelodyModel, helpers
```
Wuestengecko, line 8: The `MelodyModel` import is no longer used in this test module.

```suggestion
from capellambse import helpers
```

Wuestengecko (changes requested):
Looking very good! Only some minor nitpicks.

ewuerger:
> As the filter names in the XML are quite cryptic and arbitrary at times, it might be a good idea to add an Enum or something that has all the filters that exist. Each member could have a short, but ideally still readable name, and map to the internal XML name. Then each docstring could have the full name shown in the GUI.

I added the [`_enum.py`](https://github.com/DSD-DBS/py-capellambse/pull/84/commits/53d5ff8d92f4e94dc41443c4b7d119cfaadc4c7d#diff-53b1e80fa82574ef23290a955df05ed180a46b18d6768c220b3ee1a59256a7b1R13) where all found filters are stored. Let's finalize this PR by using the filter names defined there. What do you think @Wuestengecko?

PR #94, opened 2022-06-03
Support for FunctionalChainInvolvement context and exchanged items

### LINKED ISSUE #92: Add support for functional chain involvement context
While modeling processes (operational, functional) people capture context information which is attached to exchanges (i.e. configuration / values of an exchange item, constraints, etc). In the current API via `.involved` property we skip the involvement object as technical and deliver a list of end elements (functions or operational activities, functional exchanges, etc) leaving no means for working with the involvement context itself. What would help is a new property of a functional chain / operational process like `.involvements` that would deliver a list of involvement objects where the end user could retrieve the involvement context or involved element.

Capella metamodel analysis follows:
image

image

image

The below view is of particular interest for the usecase as it shows the `exchange_context` relationship with `Constraint`:

image

ewuerger:
This PR fixes #92. In doing so we deleted the `involved` attribute on the `OperationalProcess` and made it inheriting from `fa.FunctionalChain` (look [here](https://github.com/DSD-DBS/py-capellambse/compare/functional-chain-involvment-context?expand=1#diff-dfad51eda03198b896fddb30c1477524aa68e8afb522585c75a561b90b2c6750L65-L73)). The `involved` attribute comes from the super class.

This is solely to stay dry when implementing functionality of FunctionalChains/OperationalProcesses.

### DIFF
--- a/capellambse/model/crosslayer/fa.py
+++ b/capellambse/model/crosslayer/fa.py
@@ -21,7 +21,7 @@ from capellambse.loader import xmltools
 
 from .. import common as c
 from .. import modeltypes
-from . import capellacommon, information
+from . import capellacommon, capellacore, information
 
 if t.TYPE_CHECKING:
     from . import cs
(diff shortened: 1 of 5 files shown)


capellambse/model/crosslayer/fa.py
```diff
@@ -150,10 +150,39 @@ def owner(self) -> ComponentExchange | None:
...
+        information.ExchangeItem, "exchangedItems", aslist=c.ElementList
+    )
+    exchange_context = c.AttrProxyAccessor(
+        capellacore.Constraint, "exchangeContext"
+    )
+    involved = c.AttrProxyAccessor(c.GenericElement, "involved")
+
+    @property
+    def name(self) -> str:  # type: ignore
+        return f"[{self.__class__.__name__}] to {self.involved.name}"
```
Wuestengecko, line 167: Names are not necessarily unique, it might be good to show the involved's UUID here as well just to be sure. (Or you could use the new `_short_repr_()`.)

tests/test_model_layers.py
```diff
@@ -490,6 +490,45 @@ def test_CommunicationMean(model: capellambse.MelodyModel) -> None:
...
+        pytest.param(
+            "dfc4341d-253a-4ae9-8a30-63a9d9faca39",
+            "7311e58e-9500-4c72-9620-aac32e4a1458",
+            "1a414995-f4cd-488c-8152-486e459fb9de",
+            id="FunctionalChain",
+        ),
+    ],
+)
+def test_FunctionalChainInvolvementLink_attributes(
+    model_5_2: capellambse.MelodyModel,
```
Wuestengecko, line 511: IMO it would be better to add this to the 5.0 test model, instead of adding some cases to the 5.2 one. Keeping as much as possible in the same model will make it easier to migrate to a higher minimum supported version in the future, because there's only one model with actual useful content.

ewuerger: Tbh: Whenever I test if capellambse works on a new Capella model (for e.g. 6.0) I migrate the TestModel from the latest version I used before. So all of the content from 5.0 is in 5.2 and ongoing. I don't know how much stuff is missing in the 5.0 TestModel, but I know that I (and probably we) don't want to many test models in this repo. Personally I'd like to only have one TestModel for the latest stable Capella version.

PR #97, opened 2022-06-08
Add traces attribute to GenericElement

ewuerger:
This PR adds a new ModelObject called `GenericTrace` that can be added via the Wizards > Trace Manager option exposed when right clicking a ModelObject in the Project Explorer. 
image

As this is a very generic thing, an Acessor is set on the `GenericElement` and reveals a `.traces` attribute. We currently do not distinguish between incoming and outgoing traceability links.

@Wuestengecko I want to point out that the name property of this class is in a similar fashion like the recently added `FunctionalChainInvolvementLink` & `FunctionalChainInvolvementFunction`. Does it make sense to generalize this, to get drier or do you feel that this adds unnecessary complexity? I can think about a flexible function that gets a parameter `attr: str` and builds a name just like this.

### DIFF
--- a/tests/data/melodymodel/5_2/Melody Model Test.capella
+++ b/tests/data/melodymodel/5_2/Melody Model Test.capella
@@ -1379,6 +1379,20 @@ The predator is far away</bodies>
                   sourceElement="#0fef2887-04ce-4406-b1a1-a1b35e1ce0f3"/>
             </ownedClasses>
           </ownedDataPkgs>
+          <ownedDataPkgs xsi:type="org.polarsys.capella.core.data.information:DataPkg"
+              id="80b36e27-cf57-44e6-862b-37f52b77bcc8" name="TestTrace">
+            <ownedClasses xsi:type="org.polarsys.capella.core.data.information:Class"
+                id="ed272baf-43f2-4fa1-ad50-49c00563258b" name="Class Trace"/>
+            <ownedClasses xsi:type="org.polarsys.capella.core.data.information:Class"
+                id="ad876857-33d3-4f2e-9fe2-71545a78352d" name="Class 2">
+              <ownedTraces xsi:type="org.polarsys.capella.core.data.capellacommon:GenericTrace"
+                  id="9f84f273-1af4-49c2-a9f1-143e94ab816b" targetElement="#ed272baf-43f2-4fa1-ad50-49c00563258b"
+                  sourceElement="#ad876857-33d3-4f2e-9fe2-71545a78352d"/>
+              <ownedTraces xsi:type="org.polarsys.capella.core.data.capellacommon:GenericTrace"
+                  id="0880af85-4f96-4a77-b588-2e7a0385629d" targetElement="#01788b49-ccef-4a37-93d2-119287f8dd53"
+                  sourceElement="#ad876857-33d3-4f2e-9fe2-71545a78352d"/>
+            </ownedClasses>
+          </ownedDataPkgs>
           <ownedClasses xsi:type="org.polarsys.capella.core.data.information:Class"
               id="959b5222-7717-4ee9-bd3a-f8a209899464" name="1st Specialization of SuperClass"
               summary="" description="&lt;p>This is a test description for a class.&lt;/p>&#xA;"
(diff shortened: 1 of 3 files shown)


tests/data/melodymodel/5_2/Melody Model Test.capella
```diff
@@ -1379,6 +1379,20 @@ The predator is far away</bodies>
                   sourceElement="#0fef2887-04ce-4406-b1a1-a1b35e1ce0f3"/>
             </ownedClasses>
           </ownedDataPkgs>
+          <ownedDataPkgs xsi:type="org.polarsys.capella.core.data.information:DataPkg"
+              id="80b36e27-cf57-44e6-862b-37f52b77bcc8" name="TestTrace">
+            <ownedClasses xsi:type="org.polarsys.capella.core.data.information:Class"
+                id="ed272baf-43f2-4fa1-ad50-49c00563258b" name="Class Trace"/>
```
Wuestengecko, line 1385: This name is a bit confusing, this isn't a trace by itself but rather a class that acts as target for a trace.

tests/test_model_layers.py
```diff
@@ -530,6 +530,31 @@ def test_FunctionalChainInvolvementLink_attributes(
...
+def test_Class_has_GenericTraces(
+    model_5_2: capellambse.MelodyModel, trace_uuid: str, target_uuid: str
+) -> None:
+    cls = model_5_2.by_uuid("ad876857-33d3-4f2e-9fe2-71545a78352d")
+    trace = model_5_2.by_uuid(trace_uuid)
+    target = model_5_2.by_uuid(target_uuid)
+    expected_end = f"{target.name} ({target.uuid})"
+
+    assert trace in cls.traces
+    assert trace.name == f"[GenericTrace] to {expected_end}"
```
Wuestengecko, line 555: I'm not entirely sure whether it makes sense to test this property, because it's auto-generated. This essentially means that we test the code that generates this property with code that generates this property ... which is redundant.

It'd be a different matter if the generated string was part of the test parameters, because we wouldn't duplicate the code-under-test in the test itself.

Wuestengecko (changes requested):
This looks very nice and clean.

---

> I want to point out that the name property of this class is in a similar fashion like the recently added `FunctionalChainInvolvementLink` & `FunctionalChainInvolvementFunction`. Does it make sense to generalize this, to get drier or do you feel that this adds unnecessary complexity?

Adding a function that abstracts `return f"[{self.__class__.__name__}] to {self.involved.name} ({self.involved.uuid})"` very much sounds like unnecessary complexity for me. Unless you plan for it to do significantly more, but that begs the question what else it *could* do.

Instead of abstracting it away - does it even make sense to *have* this generated property in the first place? The original purpose of `capellambse` is to access things that exist in the XML and manipulate those. Generated properties neither exist (directly) in the XML, nor can be manipulated. And additionally, everything that `name` shows can be obtained from elsewhere anyways. (I suppose the same question would also apply to the `FunctionalChainInvolvement` and related classes, as you hinted at in your OP.)

ewuerger:
> This looks very nice and clean.
Thank you!
> [...]
> [...] The original purpose of `capellambse` is to access things that exist in the XML and manipulate those. Generated properties neither exist (directly) in the XML, nor can be manipulated. And additionally, everything that `name` shows can be obtained from elsewhere anyways. (I suppose the same question would also apply to the `FunctionalChainInvolvement` and related classes, as you hinted at in your OP.)

Yes. Headless access and manipulation of model data is the main use case for capellambse. I was convinced that the generated properties should reflect the information that is displayed in Capella as close or even in an optimized way (we display the UUIDs too, such that exact matching is possible). This makes life of users which (I suppose) want to work with the data displayed in Capella much easier, since they don't need to build the name by themselves. Pretty nice of us indeed. Does this make sense to you?

Since I smell a pattern here, my immediate itch is to abstract this.

tests/test_model_layers.py
ewuerger: I want to show how the trace name looks like. When I'm unsure about how things work in an external library I check if there is a test which displays how the devs intend to do it. This test should show that we are covering the same name as displayed in Capella... the dynamic `expected_end` is there to be save for changes. But you are probably right on changing this.

PR #98, opened 2022-06-08
Add FunctionalChainInvolvementFunction

ewuerger:
PR #94 is missing `FunctionalChainInvolvementFunction`s. This is fixed in this PR.

### DIFF
--- a/tests/test_model_layers.py
+++ b/tests/test_model_layers.py
@@ -513,6 +513,24 @@ def test_FunctionalChainInvolvementLink_attributes(
     link_uuid: str,
     target_uuid: str,
 ) -> None:
+    """
+    Test if a FunctionalChainLink covers all of its attributes.
+
+    Covered attributes are
+        * exchanged_items - List of all ExchangeItems spec. on the
+          FunctionalChainLink
+        * exchange_context - A context markup.
+
+    Parameters
+    ----------
+    fc_uuid
+        UUID of the FunctionalChain (i.e. parent of FunctionalChainLink)
+    link_uuid
+        UUID of the FunctionalChainLink
+    target_uuid
+        UUID of the FunctionalChainLink target (i.e. the element behind
+        `.involved`)
+    """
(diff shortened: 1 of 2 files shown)


tests/test_model_layers.py
```diff
@@ -513,6 +513,24 @@ def test_FunctionalChainInvolvementLink_attributes(
...
+
+    Parameters
+    ----------
+    fc_uuid
+        UUID of the FunctionalChain (i.e. parent of FunctionalChainLink)
+    link_uuid
+        UUID of the FunctionalChainLink
+    target_uuid
+        UUID of the FunctionalChainLink target (i.e. the element behind
+        `.involved`)
```
ewuerger, line 532: I was confused what these UUIDs were about. Yes after less than a week. So I thought adding a docstring is a good thing.

tests/test_model_layers.py
```diff
@@ -513,6 +513,24 @@ def test_FunctionalChainInvolvementLink_attributes(
     link_uuid: str,
     target_uuid: str,
 ) -> None:
+    """
+    Test if a FunctionalChainLink covers all of its attributes.
```
Wuestengecko, line 517:
```suggestion
    """Test if a FunctionalChainLink covers all of its attributes.
```

tests/test_model_layers.py
```diff
@@ -513,6 +513,24 @@ def test_FunctionalChainInvolvementLink_attributes(
...
+
+    Covered attributes are
+        * exchanged_items - List of all ExchangeItems spec. on the
+          FunctionalChainLink
+        * exchange_context - A context markup.
```
Wuestengecko, line 522: Do we really want this information duplicated here? IMO this part of the docstring doesn't add any value, but does add maintenance overhead. We could perhaps make it more clear from the code by restructuring it a bit though.

ewuerger: You are right. No we don't.

PR #104, opened 2022-07-08
Add PVMT intro notebook

ewuerger:
[Viktor raised awareness](https://www.linkedin.com/feed/update/urn:li:activity:6950609608168132608?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A6950609608168132608%2C6950917661648166912%29) for the PVMT extension of py-capellambse. I felt the urge to make an easy to understand intro notebook for accessing property values. Here not for the case of physical cables.

## Goals for this PR
Provide showcase example notebook that
- [x] Documents access and
- [ ] writing in `example/08 Introduction to PVMT.ipynb` notebook

### Access
- [x] Provide access of property value package definitions directly under the `MelodyModel`
- [ ] Update access of pvmt attributes to meet convention instead of using artificial `pvext.get_element_pv` function
- [ ] Keep validation intact

### Writing/Modifying property values
- [ ] Provide creation of relevant PVMT model elements


capellambse/extensions/pvmt.py
```diff
@@ -36,7 +36,7 @@ def from_model(
         cls, model: capellambse.MelodyModel, element: etree._Element
     ) -> PropertyValueProxy:
         """Create a PropertyValueProxy for an element."""
-        if not hasattr(model, "_pvext") or model._pvext is None:
+        if not hasattr(model, "pvext") or model.pvext is None:
```
Wuestengecko, line 39: We briefly had touched upon this point before, but here it is in writing:

IMO this shouldn't be made part of the public API, instead we should improve the public API that we already have (i.e. the classes around `PropertyValueProxy`). The object behind `_pvext` has a very different API philosophy, to the point where one could even argue to get rid of it entirely and reimplement it based purely on the `capellambse.model` classes.

examples/08 Introduction to PVMT.ipynb
```diff
@@ -0,0 +1,330 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Introduction to Property Value Management (PVMT)\n",
+    "\n",
+    "This notebook present the current capabilities of the PVMT extension for py-capellambse.\n",
```
Wuestengecko, line 9:
```suggestion
    "This notebook presents the current capabilities of the PVMT extension for py-capellambse.\n",
```

"He, she, it - das 's' muss mit." :)

examples/08 Introduction to PVMT.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "for grp in darkmagic_domain.groups:\n",
+    "    display(grp.properties)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The domain scope (i.e. what ArchitectureLayer) this extension can be apploed to is not yet processed further. Scopes have rules underneath to control on which ModelObject's this can be applied to.\n",
+    "This is not yet supported."
```
Wuestengecko, line 92: > This is not yet supported.

It should be, just take a look at all the code in `capellambse.pvmt.validation`. What leads you to believe it isn't supported at all?

examples/08 Introduction to PVMT.ipynb
```diff
@@ -0,0 +1,330 @@
...
+   "source": [
+    "dumbledore = model.la.all_components.by_name(\"Prof. A. P. W. B. Dumbledore\")\n",
+    "dumbledore.pvmt[\"DarkMagic.Power Level.Skill\"]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In our test model we have more famous wizzards. Let us define a function to collect all property values and load them into a pandas.DataFrame:"
```
Wuestengecko, line 157:
```suggestion
    "In our test model we have more famous wizards. Let us define a function to collect all property values and load them into a pandas.DataFrame:"
```

From [Wiktionary](https://en.wiktionary.org/wiki/wizzard):

> **wizzard** (*plural* [wizzards](https://en.wiktionary.org/wiki/wizzards#English))
> 1. *Obsolete spelling of [wizard](https://en.wiktionary.org/wiki/wizard#English)*

We have enough obsolete things in our code already, let's try to not introduce more. ;)

tests/data/melodymodel/5_2/Melody Model Test.capella
```diff
@@ -1,3124 +1,1612 @@
...
...org/capella/core/cs/5.0.0" xmlns:org.polarsys.capella.core.data.ctx="http://www.polarsys.org/capella/core/ctx/5.0.0" xmlns:org.polarsys.capella.core.data.epbs="http://www.polarsys.org/capella/core/epbs/5.0.0" xmlns:org.polarsys.capella.core.data.fa="http://www.polarsys.org/capella/core/fa/5.0.0" xmlns:org.polarsys.capella.core.data.information="http://www.polarsys.org/capella/core/information/5.0.0" xmlns:org.polarsys.capella.core.data.information.datatype="http://www.polarsys.org/capella/core/information/datatype/5.0.0" xmlns:org.polarsys.capella.core.data.information.datavalue="http://www.polarsys.org/capella/core/information/datavalue/5.0.0" xmlns:org.polarsys.capella.core.data.interaction="http://www.polarsys.org/capella/core/interaction/5.0.0" xmlns:org.polarsys.capella.core.data.la="http://www.polarsys.org/capella/core/la/5.0.0" xmlns:org.polarsys.capella.core.data.oa="http://www.polarsys.org/capella/core/oa/5.0.0" xmlns:org.polarsys.capella.core.data.pa="http://www.polarsys.org/capella/core/pa/5.0.0" xmlns:org.polarsys.capella.core.data.pa.deployment="http://www.polarsys.org/capella/core/pa/deployment/5.0.0" id="af2196ac-49d3-4063-885c-9fa29adc39a8" name="Melody Model Test">
```
Wuestengecko, line 4: The formatting in this file looks suspicious - it's neither Capella nor our EXS module nor pure LXML. What happened here?

I don't like having this kind of reformat-a-huge-file-change in a small, feature focused branch/PR like this, it makes reviewing the few actual changes just so much harder.

tests/data/melodymodel/5_2/Melody Model Test.capella
```diff
@@ -1,3124 +1,1612 @@
...
-            description="&lt;p>Principal of Hogwarts, wearer of the elder wand and greatest mage of all time.&lt;/p>&#xA;"
-            actor="true" human="true">
-          <ownedFunctionalAllocation xsi:type="org.polarsys.capella.core.data.fa:ComponentFunctionalAllocation"
-              id="fac5ccbe-00d1-4f0a-8cbc-781330113267" targetElement="#f708bc29-d69f-42a0-90cc-11fc01054cd0"
-              sourceElement="#08e02248-504d-4ed8-a295-c7682a614f66"/>
-          <ownedFunctionalAllocation xsi:type="org.polarsys.capella.core.data.fa:ComponentFunctionalAllocation"
-              id="4202800b-4058-4075-9df5-59cc2ca92e3d" targetElement="#beaf5ba4-8fa9-4342-911f-0266bb29be45"
-              sourceElement="#08e02248-504d-4ed8-a295-c7682a614f66"/>
-          <ownedFeatures xsi:type="org.polarsys.capella.core.data.fa:ComponentPort"
-              id="9dcf355a-84a7-44df-a480-8911a85fd0cd" name="CP 1" orientatio
```
Wuestengecko, line 1396: Typo: Voldemeort ⇒ Voldemort

examples/08 Introduction to PVMT.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "display(darkmagic_domain.groups)\n",
+    "for grp in darkmagic_domain.groups:\n",
+    "    display(grp.properties)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The domain scope (i.e. what ArchitectureLayer) this extension can be apploed to is not yet processed further. Scopes have rules underneath to control on which ModelObject's this can be applied to.\n",
```
Wuestengecko, line 91:
```suggestion
    "The domain scope (i.e. what ArchitectureLayer) this extension can be applied to is not yet processed further. Scopes have rules underneath to control on which ModelObject's this can be applied to.\n",
```

Wuestengecko (changes requested):

capellambse/model/crosslayer/capellacore.py
```diff
@@ -21,11 +24,153 @@ class Constraint(c.GenericElement):
...
+            for scope_type, scope in group_scopes:
+                if scope_type == "ARCHITECTURE":
+                    return scope.split(";") if ";" in scope else scope
+
+        return self.description
+
+
+@c.xtype_handler(None)
+class PropertyValuePkg(c.GenericElement):
+    """A Package for PropertyValues also called ``Domain``."""
```
Wuestengecko, line 137: > also called `Domain`

Is it really? I think the standard `PropertyValuePkg` which this class represents and the "Domain" from the property value *management* extension are two different things (the latter expanding upon the former).

capellambse/model/crosslayer/capellacore.py
```diff
@@ -21,11 +24,153 @@ class Constraint(c.GenericElement):
...
+    @property
+    def scope(self) -> str | list[str] | None:
+        """Valid architecture layer(s) for this ``PropertyValueGroup``."""
+        if self.description is not None:
+            group_scopes = pvmt.validation.yield_group_scope(self.description)
+            for scope_type, scope in group_scopes:
+                if scope_type == "ARCHITECTURE":
+                    return scope.split(";") if ";" in scope else scope
+
+        return self.description
```
Wuestengecko, line 132: The "scope" is a) much more than just the valid architecture layers (it encompasses all validation rules, and as a user I would expect to be able to access and modify all of them) and b) defined by the PVMT extension (as far as I'm aware), and should therefore be part of `capellambse.extensions.pvmt` instead.

capellambse/model/crosslayer/capellacore.py
ewuerger: I'm convinced it is. `PropertyValuePkg` has a self reference such that child `PropertyValuePkg`s can be interpreted as sub(sub/subsub...) domains.?

capellambse/model/crosslayer/capellacore.py
```diff
@@ -21,11 +24,153 @@ class Constraint(c.GenericElement):
...
+            BooleanPropertyValue,
+            IntegerPropertyValue,
+            FloatPropertyValue,
+            EnumerationPropertyValue,
+        ),
+        aslist=c.ElementList,
+    )
+
+    @property
+    def scope(self) -> str | list[str] | None:
```
Wuestengecko, line 124: This function can return one of three types, and I don't see a reason to overcomplicate it like this.

- Instead of a single `str`, return a list with one element
- Instead of `None`, return an empty list

Removes a lot of complexity without losing any value, as you can still easily distinguish between the three cases based on the `len()` of the returned list if you need to.

PR #109, opened 2022-07-20
Minor fixes

ewuerger:
With this PR we are more safe during loader failures and the displayed representation of `reqif.ReqIFElement`s in the Terminal (and possibly Jupyter) is fixed. For the latter we simply had to add a `_short_repr_`. To stay dry we changed the signature of `_short_repr_` on `common.GenericElement` such that it can be reused.

### DIFF
--- a/capellambse/extensions/reqif/elements.py
+++ b/capellambse/extensions/reqif/elements.py
@@ -312,12 +312,11 @@ class AbstractRequirementsAttribute(c.GenericElement):
     definition = c.AttrProxyAccessor(AttributeDefinition, "definition")
 
     def __repr__(self) -> str:
-        mytype = self.xtype.rsplit(":", maxsplit=1)[-1]
-        try:
-            name = self.definition.long_name
-        except AttributeError:
-            name = ""
-        return f"<{mytype} [{name}] ({self.uuid})>"
+        return self._short_repr_()
+
+    def _short_repr_(self, _: str = "") -> str:
+        name = self.definition.long_name if self.definition is not None else _
+        return super()._short_repr_(name)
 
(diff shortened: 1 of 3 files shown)


capellambse/extensions/reqif/elements.py
```diff
@@ -279,6 +279,9 @@ def __repr__(self) -> str:  # pragma: no cover
 
         return f'<{mytype} {"/".join(reversed(path))!r} ({self.uuid})>'
 
+    def _short_repr_(self, name: str = "") -> str:
+        return super()._short_repr_(name or self.name or self.long_name)
```
Wuestengecko, line 283: Why not just `return f"<{type(self).__name__} {self.name or self.long_name!r} ({self.uuid})>"` here? No need to overcomplicate the superclass to accomodate for an odd subclass. That's what inheritance is about after all. And accepting a `name` from externally into `_short_repr_` feels just ... wrong. It's not an internal function, so the universe can call it - we really need to take better care of our public API.

Wuestengecko (changes requested):

Wuestengecko:
> Ran into a situation where stderr was already a string.

I'm curious about that "situation". How did that happen? Can you post some code?

capellambse/extensions/reqif/elements.py
ewuerger: The `super()._short_repr_(name or self.name or self.long_name)` does exactly what you ask for. Why not reuse this from the super class? I thought that is what inheritance is about. It is also dry.
Am I misunderstanding that `_short_repr_` is an internal function which is only ever called by `__repr__` from `GenericElement` or `ElementList`. Never ever, by the user directly. That's why `repr()` exists.

PR #110, opened 2022-07-21
Implement `_repr_mimebundle_` on diagrams in `capellambse.model`

Wuestengecko:
The unique situation with the diagram cache requires a bit more delicate control over the MIME types handed to the front-end for displaying a diagram than is possible with dedicated methods per MIME type. The MIME bundle approach allows us to exert that control. This PR fixes a concrete issue with cache handling, which was the original reason for implementing this change.

It is possible to have a cache for one image format, but not others (e.g. cached SVGs, but not PNGs). In this case, for the non-cached image formats, the diagram would be rendered by the `capellambse.aird` module. This could lead to differences in the displayed diagram depending on which format the frontend selected.

Instead, if a cache is available and at least one of the allowed formats in the MIME bundle is cached, then only cached data will be returned. If no format can be loaded from cache, or no cache is available in the first place, the supported formats will be rendered by `capellambse` as usual.

Additionally, as now every registered DiagramConverter can specify its own MIME type, the possible formats delivered to the frontend are no longer limited by the methods implemented on the diagram classes.


capellambse/model/diagram.py
```diff
@@ -134,13 +134,57 @@ def _short_html_(self) -> markupsafe.Markup:
...
+        if include is None:
+            include = helpers.EverythingContainer()
+        if exclude is None:
+            exclude = ()
+
+        formats: dict[str, DiagramConverter] = {}
+        eligible_formats = [("image/png", "png")]
+        # XXX Hack to disable SVG in Jupyter notebooks, when requested
+        if not self._model.jupyter_untrusted:
+            eligible_formats.append(("image/svg+xml", "svg"))
```
ewuerger: Why is it a hack? Is `_repr_mimebundle_` also not an internal method or appearing in the documentation?

ewuerger (approved):
I suppose you tested it in a jupyter notebook.
Then good job: LGTM.

capellambse/model/diagram.py
Wuestengecko: grafik

`_repr_mimebundle_` is a method [from the IPython kernel](https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject._repr_mimebundle_) (the docu sucks, but here's an [example notebook](https://github.com/ipython/ipython/blob/master/examples/IPython%20Kernel/Custom%20Display%20Logic.ipynb) that elaborates a bit more).

We simply implement this method, so that Jupyter displays the diagrams nicely without needing any special function calls around them (i.e. so you can just say `diagram` in a cell instead of e.g. `SVG(diagram.as_svg)`).

Before this, we had the `_repr_svg_` and `_repr_png_` methods, which did essentially the same thing, albeit in a slightly different order than `_repr_mimebundle_` does now. And the `_repr_svg_` was "hacked" the same way to disable it when `jupyter_untrusted` was set, it just wasn't explicitly marked as being a hack like it is now.

ewuerger (approved):

ewuerger (approved):
LGTM

PR #112, opened 2022-07-22
ReqIF Attribute Value default __repr__ behaviour

ewuerger:
When there is no definition reference given, Capella displays AttributeValue elements with `[TYPE Value Attribute]` where the `TYPE` is exchanged with one of (Integer, String, Boolean,...). This commit implements this behavior and adds lines to `test_create_requirements_attributes`.

### DIFF
--- a/capellambse/extensions/reqif/elements.py
+++ b/capellambse/extensions/reqif/elements.py
@@ -59,6 +59,7 @@ __all__ = [
 import collections.abc as cabc
 import logging
 import os
+import re
 import typing as t
 
 import markupsafe
@@ -311,13 +312,17 @@ class AbstractRequirementsAttribute(c.GenericElement):
 
     definition = c.AttrProxyAccessor(AttributeDefinition, "definition")
 
+    value: xmltools.AttributeProperty | c.AttrProxyAccessor
+
     def __repr__(self) -> str:
         return self._short_repr_()
 
     def _short_repr_(self) -> str:
+        mytype = type(self).__name__
         if self.definition is not None:
-            return f"<{type(self).__name__} {self.definition.long_name!r} ({self.uuid})>"
-        return super()._short_repr_()
+            return f"<{mytype} {self.definition.long_name!r} ({self.uuid})>"
+        default_name = re.sub(r"([A-Z])", r" \1", mytype).split()
+        return f"<{mytype} [{' '.join(default_name)}] ({self.uuid})>"
 
     def _short_html_(self) -> markupsafe.Markup:
         if self.definition is not None:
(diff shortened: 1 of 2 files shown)


Wuestengecko (changes requested):
There's a bunch of ~~already-merged changes~~ unrelated changes from a different branch mixed into this PR, which makes reviewing it unnecessarily difficult. Can you please rebase only the commits that you meant to merge on top of latest `master`?

capellambse/extensions/reqif/elements.py
```diff
@@ -311,13 +312,17 @@ class AbstractRequirementsAttribute(c.GenericElement):
...
         return self._short_repr_()
 
     def _short_repr_(self) -> str:
+        mytype = type(self).__name__
         if self.definition is not None:
-            return f"<{type(self).__name__} {self.definition.long_name!r} ({self.uuid})>"
-        return super()._short_repr_()
+            return f"<{mytype} {self.definition.long_name!r} ({self.uuid})>"
+        default_name = re.sub(r"([A-Z])", r" \1", mytype).split()
+        return f"<{mytype} [{' '.join(default_name)}] ({self.uuid})>"
```
Wuestengecko, line 325: We can avoid the split-rejoin by slightly altering the regexp:

```suggestion
        default_name = re.sub(r"(?<!^)([A-Z])", r" \1", mytype)
        return f"<{mytype} [{default_name}] ({self.uuid})>"
```

I `timeit`'d it, and there isn't actually a difference in run time between the two (unlike what I expected). So whether you take it or not is up to you.

tests/test_reqif.py
```diff
@@ -440,27 +440,29 @@ def test_create_requirements_attributes_with_default_values(
...
-            pytest.param("Float", 0.0, None),
-            pytest.param("Float", 1.0, "1.0"),
-            pytest.param("Date", None, None),
+            pytest.param("Int", 0, None, "[Integer Value Attribute]"),
+            pytest.param("Int", 1, "1", "[Integer Value Attribute]"),
+            pytest.param("Str", "", None, "[String Value Attribute]"),
+            pytest.param("Str", "test", "test", "[String Value Attribute]"),
+            pytest.param("Float", 0.0, None, "[Real Value Attribute]"),
+            pytest.param("Float", 1.0, "1.0", "[Real Value Attribute]"),
+            pytest.param("Date", None, None, "[Date Value Attribute]"),
```
Wuestengecko, line 451: The `expected_repr_name` only correlates with the `type_hint`, and is completely independent from the other two parameterized arguments. It might be worth to get it out into its own test case, to prevent even more repetition from creeping in in the future.

And once we have that, we can even go and test the longer `type_hint`s now. :)

Wuestengecko (changes requested):

PR #117, opened 2022-07-28
Fix issues introduced with #116

Wuestengecko:
PR #116 ("Improve handling of HTML attributes in the model") introduced a few issues that went undetected by the tests and review. This PR fixes them.

### DIFF
--- a/capellambse/loader/xmltools.py
+++ b/capellambse/loader/xmltools.py
@@ -87,7 +87,7 @@ class AttributeProperty:
         """Generate the qualified name of this descriptor."""
         if self.__objclass__ is None:
             return f"(unknown {type(self).__name__} - call __set_name__)"
-        return f"{type(self.__objclass__.__name__)}.{self.__name__}"
+        return f"{self.__objclass__.__name__}.{self.__name__}"
 
     @t.overload
     def __get__(self, obj: None, objtype: type) -> AttributeProperty:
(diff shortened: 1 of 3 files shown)


capellambse/loader/xmltools.py
```diff
@@ -87,7 +87,7 @@ def _qualname(self) -> str:
         """Generate the qualified name of this descriptor."""
         if self.__objclass__ is None:
             return f"(unknown {type(self).__name__} - call __set_name__)"
-        return f"{type(self.__objclass__.__name__)}.{self.__name__}"
```
ewuerger, line 90: My bad should've seen that during review.

capellambse/model/diagram.py
```diff
@@ -372,6 +368,12 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
         return self._model is other._model and self._element == other._element
 
+    @property
+    def description(self) -> str:
+        """Return the diagram description."""
+        desc = self._model._loader[self.uuid].get("documentation")
+        return desc and helpers.repair_html(desc)
```
ewuerger, line 375: I'm curious on why this change is needed. The way via xmltools.HTMLAttributeProperty is not sufficient?

ewuerger (approved):
LGTM

capellambse/model/diagram.py
Wuestengecko: No, these properties can't do something as complex as `obj._model._loader[obj.uuid]`. So we need to keep doing it this way and wait until the big refactor. :(

PR #118, opened 2022-07-29
Add RequirementTypesFolder accessor to RequirementsModule

ewuerger:
The `RequirementTypesFolder` can't be created underneath a `RequirementsModule`. But it can be moved there via the GUI. In this PR we added an Accessor to the `RequirementsModule` for `RequirementTypesFolder`s.


Wuestengecko (approved):
Looks good! LGTM

PR #124, opened 2022-08-12
Minor improvements to create api

ewuerger:
See the title.

### DIFF
--- a/capellambse/model/common/accessors.py
+++ b/capellambse/model/common/accessors.py
@@ -79,10 +79,8 @@ class Accessor(t.Generic[T], metaclass=abc.ABCMeta):
 class WritableAccessor(Accessor[T], metaclass=abc.ABCMeta):
     """An Accessor that also provides write support on lists it returns."""
 
-    __slots__ = ()
-
     aslist: type[ElementListCouplingMixin] | None
-    single_attr: str | None = None
+    single_attr: str | None
 
     def __init__(
         self,
@@ -92,7 +90,7 @@ class WritableAccessor(Accessor[T], metaclass=abc.ABCMeta):
         **kw: t.Any,
     ) -> None:
         super().__init__(*args, **kw)  # type: ignore[call-arg]
-        self.single_attr = single_attr  # type: ignore[misc]
+        self.single_attr = single_attr
         if aslist is not None:
             self.aslist = type(  # type: ignore[misc]
                 "Coupled" + aslist.__name__,
@@ -131,7 +129,10 @@ class WritableAccessor(Accessor[T], metaclass=abc.ABCMeta):
     ) -> T:
         """Create and return a new single attribute element."""
         if self.single_attr is None:
-            raise TypeError("Cannot create object. You need to pass a dict.")
+            raise TypeError(
+                "Cannot create object: "
+                "You need to pass a dict where the key value pair specifies the attribute name and value."
+            )
         return self.create(elmlist, **{self.single_attr: arg})
 
     def insert(
@@ -285,7 +286,7 @@ class PhysicalAccessor(Accessor[T]):
 class DirectProxyAccessor(WritableAccessor[T], PhysicalAccessor[T]):
     """Creates proxy objects on the fly."""
 
-    __slots__ = ("follow_abstract", "rootelem", "single_attr")
+    __slots__ = ("follow_abstract", "rootelem")
 
     aslist: type[ElementListCouplingMixin] | None
     single_attr: str | None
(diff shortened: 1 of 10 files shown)


capellambse/model/common/accessors.py
```diff
@@ -82,15 +82,17 @@ class WritableAccessor(Accessor[T], metaclass=abc.ABCMeta):
...
     def __init__(
         self,
         *args: t.Any,
         aslist: type[element.ElementList] | None,
+        single_attr: str | None = None,
         **kw: t.Any,
     ) -> None:
         super().__init__(*args, **kw)  # type: ignore[call-arg]
-
+        self.single_attr = single_attr  # type: ignore[misc]
```
Wuestengecko, line 95: grafik

capellambse/model/common/accessors.py
```diff
@@ -275,9 +285,10 @@ def _make_list(self, parent_obj, elements):
 class DirectProxyAccessor(WritableAccessor[T], PhysicalAccessor[T]):
     """Creates proxy objects on the fly."""
 
-    __slots__ = ("follow_abstract", "rootelem")
+    __slots__ = ("follow_abstract", "rootelem", "single_attr")
```
Wuestengecko, line 288: The `single_attr` slot would actually belong to `WritableAccessor`, I'm not sure how much sense it makes to have it defined in all its subclasses. I'd rather not.

tests/test_model_creation_deletion.py
```diff
@@ -58,17 +64,25 @@ def test_deleted_elements_are_removed(model, deletion_target):
...
     comps = model.la.root_component.components
     assert len(comps) == 2
 
     comps.delete_all(name="Delete Me")
     assert len(comps) == 1
     assert comps[0].name == "Keep Me"
+
+
+def test_delete_elements_from_lookups(model: capellambse.MelodyModel):
+    assert model.la.all_functions[0] == model.la.root_function
```
Wuestengecko, line 84: We should rather `assert` over the entire list here, just to guard against the test model accidentally being changed in the future.

```suggestion
    assert model.la.all_functions == [model.la.root_function]
```

tests/test_reqif.py
```diff
@@ -531,6 +531,17 @@ def test_create_enum_value_attribute_on_requirements(
...
+    def test_create_enum_value_attribute_with_passing_values(
+        self, model: capellambse.MelodyModel
+    ):
+        req = model.oa.all_requirements[0]
+        dtdef = model.by_uuid("637caf95-3229-4607-99a0-7d7b990bc97f")
+
+        attr = req.attributes.create("Enum", values=dtdef.values[:1])
+
+        assert attr in req.attributes
+        assert attr.values == dtdef.values[:1]
```
Wuestengecko, line 543: This looks a bit hacky, can we not just use simple `model.by_uuid` access instead of this `.values[:1]` thing here?

It also leads to a useless test if `dtdef.values` happens to become empty for some reason (like a bug that leads to the values not being found and an empty list returned instead).

tests/test_reqif.py
```diff
@@ -639,6 +650,87 @@ def test_create_RequirementType_AttributeDefinition_creation(
...
+                    "type_hints": "EnumerationDataTypeDefinition",
+                    "long_name": "TestEnumAttrDataTypeDef",
+                    "values": ["a", "b"],
+                },
+            ],
+        )
+
+        adtdef, edtdef = reqtypesfolder.data_type_definitions
+
+        assert isinstance(adtdef, reqif.DataTypeDefinition)
```
Wuestengecko, line 728: The spacing makes it seem as if the `adtdef, edtdef = reqtypesfolder.data_type_definitions` is the line-under-test, when it's actually the `create` statement above. Judging from the test case name, this line actually belong the assertions though, and should not be separated from them by an empty line.

Wuestengecko (changes requested):

capellambse/model/common/accessors.py
ewuerger: I tried to have `single_attr` on the `WritableAccessor` where it should belong to... but Python doesn't let me because of 
`TypeError: multiple bases have instance lay-out conflict`.

Have a look at [this great answer for usage of `__slots__` and its caveats](https://stackoverflow.com/questions/472000/usage-of-slots#answer-28059785). The interesting section for our problem here is named `Biggest Caveat: Multiple inheritance`.

ewuerger: Solution: Just don't include this attribute in the slots.

capellambse/model/common/element.py
```diff
@@ -214,6 +214,36 @@ def __init__(
                     raise TypeError(
                         f"Cannot set {key!r} on {type(self).__name__}"
                     )
+                elif isinstance(val, cabc.Mapping):
+                    raise NotImplementedError
```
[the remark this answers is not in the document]
ewuerger: Added a string to inform the user for unsupported `Multicreation`.

Wuestengecko (changes requested):

Wuestengecko (approved):
I've gone ahead and added another small test case for the list item creation test in `test_reqif` - just for my own peace of mind. Looks good now, let's get this merged!

Issue #85, opened 2022-05-25
Raise minimum supported Python version to 3.9


Wuestengecko:
Keeping support for Python 3.8 introduces a notable amount of developer overhead. This is primarily due to it being the last version without [PEP-585](https://peps.python.org/pep-0585/). This PEP deprecates most of the stdlib-parallel classes in `typing` (e.g. `List`, `Set`, etc.) and allows to use type hints on the stdlib classes themselves.

This works fine in a type annotation context, however when we want to subclass something we still need to go back to the `typing` classes for that instance. This is not only annoying, but also makes the code a little less clear to read due to essentially the same class being used from different modules.

Dropping support for Python 3.8 entirely is at this point a reasonable choice, in my opinion. However, we do need to analyze our dependency chain and make sure that all dependencies and dependents, as well as all relevant tool configuration, have been updated accordingly.

Issue #96, opened 2022-06-07
Clean tests


ewuerger:
Remove unused test models and merge test models into latest version

---

IMO it would be better to add this to the 5.0 test model, instead of adding some cases to the 5.2 one. Keeping as much as possible in the same model will make it easier to migrate to a higher minimum supported version in the future, because there's only one model with actual useful content.

_Originally posted by @Wuestengecko in https://github.com/DSD-DBS/py-capellambse/pull/94#discussion_r888822913_

PR #129, opened 2022-08-16
Raise minimum supported Python version to 3.9 (#85)

Wuestengecko:
This PR also includes a general update to the project infrastructure first, which makes the project REUSE compliant among several other minor improvements.


ewuerger (approved):
Seeing this PR lets me remember that we wanted to remove some redundant (imo) test model files such that the reuse tool isn't multiplying the counter by 2. For e.g. the 1_4 (never used in any tests) and 5_2 model should be removed. I think this is the goal of #96 and there is additional work to be done for migration of data from the 5_2 back into the 5_0 model.

LGTM!

PR #130, opened 2022-08-17
feat: Add a simple REPL launcher

Wuestengecko:
This PR adds a small script which launches a simple Read-Execute-Print loop based on the stdlib `code` module. It is intended to facilitate quick and simple debugging action where it would be too much overhead to launch a full blown Jupyter server. However, it is also very useful as a quick and easy way to look up some things in a model.


ewuerger (approved):
LGTM

PR #131, opened 2022-08-17
Added missing SystemAnalysis root component symbol

ewuerger:
Literally the title.

### DIFF
--- a/tests/test_svg.py
+++ b/tests/test_svg.py
@@ -35,6 +35,7 @@ TEST_DIAGS = [
     "[OCB] Operational Capabilities",
     "[OPD] Obtain food via hunting",
     "[MSM] States of Functional Human Being",
+    "[SAB] System",
     "[OEBD] Operational Context",
     "[PAB] Physical System",
     "[LDFB] Test flow",
--- a/capellambse/svg/style.py
+++ b/capellambse/svg/style.py
@@ -61,6 +61,7 @@ STATIC_DECORATIONS: dict[str, tuple[str, ...]] = {
         "CapabilitySymbol",
         "MissionSymbol",
         "SystemActorSymbol",
+        "SystemComponentSymbol",
         "SystemHumanActorSymbol",
     ),
     "Mode State Machine": (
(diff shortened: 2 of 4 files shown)


tests/test_svg.py
```diff
@@ -35,6 +35,7 @@
     "[OCB] Operational Capabilities",
     "[OPD] Obtain food via hunting",
     "[MSM] States of Functional Human Being",
+    "[SAB] System",
```
Wuestengecko, line 38: I wonder how useful it actually is to add the new diagram to this list. The only test that uses this list (`TestSVG.test_diagram_decorations`) doesn't actually test anything about the icons/symbols, and only checks whether an exception flies ... which didn't happen before this change anyways.

In other words, if we revert the fix in production code and only keep the updated tests, they're still green. Which makes them functionally useless and therefore essentially just a waste of CPU time:

```
% pytest -q --disable-warnings tests/test_svg.py
..............sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 67%]
sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss..................                                                                                                           [100%]
32 passed, 272 skipped, 60 warnings in 1.91s
% git show @ -- capellambse |patch -Rp1
patching file capellambse/svg/style.py
patching file capellambse/svg/symbols.py
% pytest -q --disable-warnings tests/test_svg.py
..............sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 67%]
sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss..................                                                                                                           [100%]
32 passed, 272 skipped, 60 warnings in 1.91s
```

Wuestengecko: After discussing this with @ewuerger: Let's just keep the test and get this merged. Better too many tests than too few of them. :)

Wuestengecko (approved):

Issue #128, opened 2022-08-16
Exchange Scenario Diagram not working.


juanvaccarezza:
I just modify the diagram to look like this.

![image](https://user-images.githubusercontent.com/1491039/184900385-7b8c4dea-01c5-4ef8-8d54-7d595ecd7a07.png)

Then try to show it in jupyter by:

`model.diagrams.by_name("[ES] Exchange Scenario")`

I got:

```
Ignoring unknown filter type 'diagram:AbsoluteBoundsFilter'
Ignoring unknown filter type 'diagram:AbsoluteBoundsFilter'
Ignoring unknown filter type 'diagram:AbsoluteBoundsFilter'
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/.local/lib/python3.8/site-packages/IPython/core/formatters.py:973, in MimeBundleFormatter.__call__(self, obj, include, exclude)
    970     method = get_real_method(obj, self.print_method)
    972     if method is not None:
--> 973         return method(include=include, exclude=exclude)
    974     return None
    975 else:

File ~/.local/lib/python3.8/site-packages/capellambse/model/diagram.py:172, in AbstractDiagram._repr_mimebundle_(self, include, exclude, **_)
    169 if bundle:
    170     return bundle
--> 172 render = self.__render_fresh({})
    173 for mime, conv in formats.items():
    174     try:

File ~/.local/lib/python3.8/site-packages/capellambse/model/diagram.py:322, in AbstractDiagram.__render_fresh(self, params)
    319         self._render = self.__create_error_image("parse", self._error)
    321 if hasattr(self, "_error"):
--> 322     raise self._error
    323 return self._render

File ~/.local/lib/python3.8/site-packages/capellambse/model/diagram.py:316, in AbstractDiagram.__render_fresh(self, params)
    314 self.invalidate_cache()
    315 try:
--> 316     self._render = self._create_diagram(params)
    317 except Exception as err:
    318     self._error = err

File ~/.local/lib/python3.8/site-packages/capellambse/model/diagram.py:404, in Diagram._create_diagram(self, params)
    403 def _create_diagram(self, params: dict[str, t.Any]) -> aird.Diagram:
--> 404     return aird.parse_diagram(self._model._loader, self._element, **params)

File ~/.local/lib/python3.8/site-packages/capellambse/aird/parser/__init__.py:183, in parse_diagram(model, descriptor, **params)
[...]

PR #132, opened 2022-08-17
fix(aird): Fix rendering issues with "custom" `<sourceAnchors>`

Wuestengecko:
Some `<sourceAnchor>` elements have a `custom` marker appended to their coordinates. This commit fixes an exception that was thrown as a result of `capellambse` not expecting that marker.

### DIFF
--- a/capellambse/aird/parser/_edge_factories.py
+++ b/capellambse/aird/parser/_edge_factories.py
@@ -115,6 +115,8 @@ def extract_bendpoints(
             ).attrib["id"]
         except (StopIteration, KeyError):
             sourceanchor = "(0.5, 0.5)"
+        if sourceanchor.endswith(" custom"):
+            sourceanchor = sourceanchor[: -len(" custom")]
         sourceanchor = helpers.ssvparse(
             sourceanchor, float, parens="()", num=2
         )


juanvaccarezza:
I think this is related to #128

capellambse/aird/parser/_edge_factories.py
```diff
@@ -115,6 +115,8 @@ def extract_bendpoints(
             ).attrib["id"]
         except (StopIteration, KeyError):
             sourceanchor = "(0.5, 0.5)"
+        if sourceanchor.endswith(" custom"):
+            sourceanchor = sourceanchor[: -len(" custom")]
```
ewuerger, line 119: Do you have any clue/assumption what this custom is there for? I've found out that Capella uses [Draw2D](https://github.com/freegroup/draw2d) for their diagrams... but for the `geo.Point` type in Draw2D there is no custom suffix to be found. It seems that this is something specific in Capella.

ewuerger (approved):
LGTM!

capellambse/aird/parser/_edge_factories.py
Wuestengecko: I presume it has to do with objects that were modified by hand (a.k.a. *custom*ized) in the Capella GUI, but that's ... just a theory. A code theory!

ewuerger:
Solves #128.

PR #134, opened 2022-08-31
Replace `%s` and `%q` in HTTP(S) URLs

Wuestengecko:
This PR implements replacement of the special markers `%s` and `%q` in URLs requested by the `HTTPFileHandler`. They both replace the URL escaped file name, with the difference that `%q` escapes a `/`, while `%s` passes the `/` along unchanged. This allows to build much more complex request URLs than was previously possible, and should allow to connect to a much wider range of hosting services.


jamilraichouni (dismissed):
Nice :-), thanks!

jamilraichouni (approved):

PR #139, opened 2022-09-07
test: Refactor for pytest's `--import-mode=importlib`

Wuestengecko:
In a recent team meeting, we agreed to migrate to this option. Here goes capellambse.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -211,6 +211,7 @@ enable = [
 
 [tool.pytest.ini_options]
 addopts = """
+  --import-mode=importlib
   --strict-config
   --strict-markers
 """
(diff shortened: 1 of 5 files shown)


ewuerger (dismissed):
Apart from me (a VSCode user) getting spammed by an angry pylinter: 
`Attempted relative import beyond top-level packagePylint(E0402:relative-beyond-top-level)`
in `test_model_layers.py` and `test_model_loading.py` 

✨LGTM✨

Wuestengecko:
> Apart from me (a VSCode user) getting spammed by an angry pylinter: `Attempted relative import beyond top-level packagePylint(E0402:relative-beyond-top-level)` in `test_model_layers.py` and `test_model_loading.py`

That's an unfortunate side effect of pytest taking over the importing instead of using standard Python import semantics (not VSC exclusive) – I'm exploring options that don't trigger linter false-positives everywhere, but for now it works.

PR #140, opened 2022-09-08
Fix linter issues in tests

Wuestengecko:
This PR is a follow-up to #139, and fixes linter issues uncovered and introduced in the test cases.


ewuerger (approved):
I waited for this my whole life:
image
image

anyways: ✨LGTM✨

Wuestengecko:
@ewuerger 

> I waited for this my whole life:  
> image

Well, do you actually see a way to fix them properly? That was my point back then IIRC, but here I don't really see how that's possible :)

PR #142, opened 2022-09-09
Minor fixes

ewuerger:
- Standardize allocated_functions attribute of components across all layers.
- Copied styles for SystemActor and SystemHumanActor from [MCB].

### DIFF
--- a/capellambse/model/layers/la.py
+++ b/capellambse/model/layers/la.py
@@ -59,7 +59,7 @@ class LogicalComponent(cs.Component):
 
     _xmltag = "ownedLogicalComponents"
 
-    functions = c.ReferencingProxyAccessor(
+    allocated_functions = c.ReferencingProxyAccessor(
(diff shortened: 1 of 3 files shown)


capellambse/model/layers/la.py
```diff
@@ -59,7 +59,7 @@ class LogicalComponent(cs.Component):
 
     _xmltag = "ownedLogicalComponents"
 
-    functions = c.ReferencingProxyAccessor(
+    allocated_functions = c.ReferencingProxyAccessor(
```
Wuestengecko, line 62: What about people that used `functions` before? That would break with this, but c0153237d0d4ce0474f4cb74f74df211dcb5da1f, which introduced this, is not marked as "breaking change".

Wuestengecko (changes requested):
Please also clean up the branch history, it seems you accidentally pushed a local test merge.

PR #144, opened 2022-09-09
Fix side by side badges in README

ewuerger:
The status badges in the README are displayed in new lines... this can be fixed by adding `:target:`s as described in [this issue](https://github.com/badges/shields/issues/1491).


Wuestengecko (approved):
Thanks!

PR #145, opened 2022-09-15
YAML-based declarative modelling

Wuestengecko:
This PR introduces a modelling language based on a YAML format. Please refer to the added documentation for details.

Mandatory main goals for this PR:

- [x] Implement a `dump` and `load` function, which convert between Python and the YAML format
- [x] Implement the `apply` function, which consumes a YAML file and modifies a model
  - [x] Select existing model objects via `parent: !uuid 1234abcd-...`
  - [x] Create new child objects with the `create:` YAML keyword
  - [x] Modify properties of existing objects with the `modify:` YAML keyword
  - [x] Delete child objects with the `delete:` YAML keyword
- [x] Add documentation

Optional goals, which may also be moved to a separate, later PR:

- [x] Add an example notebook
- [ ] Add a YAML Schema for input validation and LSP support
- [ ] In `apply`, select existing model objects with a query string like `parent: la.root_component`

### DIFF
--- a/capellambse/decl.py
+++ b/capellambse/decl.py
@@ -1,0 +1,113 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capellambse contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Support for YAML-based declarative modelling.
+
+A YAML-based approach to describing how to create and modify
+``capellambse`` compatible models.
+"""
+from __future__ import annotations
+
+__all__ = [
+    "Promise",
+    "UUIDReference",
+    "YDMDumper",
+    "YDMLoader",
+    "dump",
+    "load",
+]
+
+import collections.abc as cabc
+import contextlib
+import dataclasses
+import os
+import typing as t
+
+import yaml
+
+from capellambse import helpers
+
+FileOrPath = t.Union[t.IO[str], str, os.PathLike[t.Any]]
+
+
+def dump(instructions: cabc.Sequence[cabc.Mapping[str, t.Any]]) -> str:
+    """Dump an instruction stream to YAML."""
+    return yaml.dump(instructions, Dumper=YDMDumper)
+
+
+def load(file: FileOrPath) -> list[dict[str, t.Any]]:
+    """Load an instruction stream from a YAML file.
+
+    Parameters
+    ----------
+    file
+        An open file-like object, or a path or PathLike pointing to such
+        a file. Files are expected to use UTF-8 encoding.
+    """
+    if hasattr(file, "read"):
+        file = t.cast(t.IO[str], file)
+        ctx: t.ContextManager[t.IO[str]] = contextlib.nullcontext(file)
+    else:
+        assert not isinstance(file, t.IO)
+        ctx = open(file, encoding="utf-8")
+
+    with ctx as opened_file:
+        return yaml.load(opened_file, Loader=YDMLoader)
+
+
+@dataclasses.dataclass(frozen=True)
+class Promise:
+    """References a model object that will be created later."""
+
+    identifier: str
+
+
+@dataclasses.dataclass(frozen=True)
+class UUIDReference:
+    """References a model object by its UUID."""
+
+    uuid: helpers.UUIDString
+
+    def __post_init__(self) -> None:
+        if not helpers.is_uuid_string(self.uuid):
+            raise ValueError(f"Malformed `!uuid`: {self.uuid!r}")
+
+
+class YDMDumper(yaml.SafeDumper):
+    """A YAML dumper with extensions for declarative modelling."""
+
+    def represent_promise(self, data: t.Any) -> yaml.Node:
+        assert isinstance(data, Promise)
+        return self.represent_scalar("!promise", data.identifier)
+
+    def represent_uuidref(self, data: t.Any) -> yaml.Node:
+        assert isinstance(data, UUIDReference)
+        return self.represent_scalar("!uuid", data.uuid)
+
+
+YDMDumper.add_representer(Promise, YDMDumper.represent_promise)
+YDMDumper.add_representer(UUIDReference, YDMDumper.represent_uuidref)
+
+
+class YDMLoader(yaml.SafeLoader):
+    """A YAML loader with extensions for declarative modelling."""
+
+    def construct_promise(self, node: yaml.Node) -> Promise:
+        if not isinstance(node, yaml.ScalarNode):
+            raise TypeError("!promise only accepts scalar nodes")
+        data = self.construct_scalar(node)
+        if not isinstance(data, str):
+            raise TypeError("!promise only accepts string scalars")
+        return Promise(data)
+
+    def construct_uuidref(self, node: yaml.Node) -> UUIDReference:
+        if not isinstance(node, yaml.ScalarNode):
+            raise TypeError("!uuid only accepts scalar nodes")
+        data = self.construct_scalar(node)
+        if not helpers.is_uuid_string(data):
+            raise ValueError(f"Not a well-formed UUID string: {data}")
+        return UUIDReference(data)
+
+
+YDMLoader.add_constructor("!promise", YDMLoader.construct_promise)
+YDMLoader.add_constructor("!uuid", YDMLoader.construct_uuidref)
--- a/capellambse/decl.py
+++ b/capellambse/decl.py
@@ -32,6 +32,7 @@ from capellambse import helpers
 from capellambse.model import common
 
 FileOrPath = t.Union[t.IO[str], str, os.PathLike[t.Any]]
+_FulfilledPromise = tuple["Promise", capellambse.ModelObject]
 
 
 def dump(instructions: cabc.Sequence[cabc.Mapping[str, t.Any]]) -> str:
@@ -80,6 +81,7 @@ def apply(model: capellambse.MelodyModel, file: FileOrPath) -> None:
     """
     instructions = load(file)
 
+    promises = dict[Promise, capellambse.ModelObject]()
     for instruction in instructions:
         parent = instruction.pop("parent")
         if isinstance(parent, UUIDReference):
@@ -89,31 +91,42 @@ def apply(model: capellambse.MelodyModel, file: FileOrPath) -> None:
                 op = instruction.pop(op_type)
             except KeyError:
                 continue
-            apply_op(parent, op)
+            for promise, object in apply_op(promises, parent, op):
+                if promise in promises:
+                    raise ValueError(
+                        f"promise_id defined twice: {promise.identifier}"
+                    )
+                promises[promise] = object
         if instruction:
             keys = ", ".join(instruction)
             raise ValueError(f"Unrecognized keys in instruction: {keys}")
 
 
 def _operate_create(
-    parent: capellambse.ModelObject, creations: dict[str, t.Any]
-):
+    promises: dict[Promise, capellambse.ModelObject],
+    parent: capellambse.ModelObject,
+    creations: dict[str, t.Any],
+) -> cabc.Generator[_FulfilledPromise, t.Any, None]:
     for attr, value in creations.items():
         if not isinstance(value, cabc.Iterable):
             raise TypeError("values below `create:*:` must be lists")
 
-        _create_complex_objects(parent, attr, value)
+        yield from _create_complex_objects(promises, parent, attr, value)
 
 
 def _operate_delete(
-    parent: capellambse.ModelObject, deletions: dict[str, t.Any]
-):
+    promises: dict[Promise, capellambse.ModelObject],
+    parent: capellambse.ModelObject,
+    deletions: dict[str, t.Any],
+) -> cabc.Generator[_FulfilledPromise, t.Any, None]:
     raise NotImplementedError("Deleting objects is not yet implemented")
 
 
 def _operate_modify(
-    parent: capellambse.ModelObject, modifications: dict[str, t.Any]
-):
+    promises: dict[Promise, capellambse.ModelObject],
+    parent: capellambse.ModelObject,
+    modifications: dict[str, t.Any],
+) -> cabc.Generator[_FulfilledPromise, t.Any, None]:
     raise NotImplementedError("Modifying objects is not yet implemented")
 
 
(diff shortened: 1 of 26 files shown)


capellambse/model/common/accessors.py
```diff
@@ -390,6 +390,32 @@ def __get__(self, obj, objtype=None):
         ]
         return self._make_list(obj, elems)
 
+    def __set__(self, obj, value):
+        is_it = isinstance(value, cabc.Iterable) and not isinstance(value, str)
```
ewuerger, line 394: You named the boolean variable `is_it` which is short for `is_iterable` I suppose. Really verbose would be `is_iterable_non_string` which reminds me of the [hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation). Something invaluable I learned during my studys (/s). There is enough space for the verbose form, so why shortening?

capellambse/model/common/element.py
```diff
@@ -214,33 +214,6 @@ def __init__(
...
-                                    type_hints = v.pop("_type")
-
-                                if not isinstance(
-                                    type_hints, cabc.Iterable
-                                ) or isinstance(type_hints, str):
-                                    type_hints = [type_hints]
-
-                                target.create(*type_hints, **v)
-                            else:
-                                target.create_singleattr(v)
```
ewuerger, line 243: Thank you for migrating this properly to `__set__` in accessors.py.

capellambse/decl.py
```diff
@@ -0,0 +1,404 @@
...
+                        raise ValueError(
+                            f"promise_id defined twice: {promise.identifier}"
+                        )
+                    promises[promise] = outcome
+                    instructions.extend(deferred.pop(promise, ()))
+        if instruction:
+            keys = ", ".join(instruction)
+            raise ValueError(f"Unrecognized keys in instruction: {keys}")
+    if deferred:
+        raise UnfulfilledPromisesError(frozenset(deferred))
```
ewuerger, line 122: This is very cool for me to read. I don't know if I shared that but I failed on implementing something with a Queue in the `apply_changeset.py`. `RuntimeError: deque mutated during iteration` >:(. I was very convinced that we'd need  a `Queue`-like structure since promises would be pushed back into it to get the proper solution. I suggest we take a closer look into special cases like re-locating `ModelObject`s and how this changes execution order of instructions. Explaining the effects of moving elements from top to bottom (in XML) and vice versa is a valuable insight to understand how the future `ChangeSet` technology and its dependency (this beautiful piece of code here) works together.

capellambse/decl.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+def _operate_create(
```
ewuerger, line 125: IMO I'd like to have this public with a section about possible raised exceptions.

Why: Give the careful user the option to read about invalid/valid input through the declarative modeling language.

capellambse/decl.py
```diff
@@ -0,0 +1,404 @@
...
+    creations: dict[str, t.Any],
+) -> cabc.Generator[_OperatorResult, t.Any, None]:
+    for attr, value in creations.items():
+        if not isinstance(value, cabc.Iterable):
+            raise TypeError("values below `create:*:` must be lists")
+
+        yield from _create_complex_objects(promises, parent, attr, value)
+
+
+def _operate_delete(
```
ewuerger, line 137: IMO I'd like to have this public with a section about possible raised exceptions.

capellambse/decl.py
```diff
@@ -0,0 +1,404 @@
...
+                raise ValueError(
+                    f"No object with UUID {obj!r} in {attr!r} of {puuid!r}"
+                ) from None
+            del uuids[idx]
+            del target[idx]
+
+    return ()
+
+
+def _operate_modify(
```
ewuerger, line 181: IMO I'd like to have this public with a section about possible raised exceptions.

capellambse/decl.py
```diff
@@ -0,0 +1,404 @@
...
+            target = getattr(parent, attr)
+        except AttributeError:
+            raise TypeError(
+                "Cannot delete object:"
+                f" {type(parent).__name__} has no attribute {attr!r}"
+            ) from None
+        if not isinstance(target, common.ElementList) or not isinstance(
+            objs, list
+        ):
+            delattr(parent, attr)
```
ewuerger, line 155: Does this `delattr` call modify the XML code? If so how? I expect some hidden execution on the `Accessor`. Also [this](https://stackoverflow.com/a/1121068/5628391) was an interesting read for me.

capellambse/decl.py
```diff
@@ -0,0 +1,404 @@
...
+
+else:
+
+    @click.command()
+    @click.option("-m", "--model", type=capellambse.ModelCLI(), required=True)
+    @click.argument("file", type=click.File("r"))
+    def _main(model: capellambse.MelodyModel, file: t.IO[str]) -> None:
+        """Apply a declarative modelling YAML file to a model."""
+        apply(model, file)
+        model.save()
```
ewuerger, line 400: Is the click CLI meant to be available optionally? This isn't documented and therefore surprises me. I'd suggest to add click as a dependency and get rid of the try/except/else routine.

ewuerger (changes requested):
Not all review comments are gate keeping. Some were opened for discussion.
After solving the important ones: ✨LGTM✨

Edit: Came back to see that I forget to tell you my whole plan I have in mind. I'm looking forward in discussing this with you. Seeing the code blocks in the great documentation you wrote I notice that we don't have a copy-button. That feels off. So we'll fix that with [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io/en/latest/). Don't rustle your jimmies yet: The reason why I annoyed you so hard on cross-referencing, is that I want to have [sphinx-hoverxef](https://sphinx-hoverxref.readthedocs.io/en/latest/) such that you don't need to click, wait (download in germany yay) and after comprehending why we send you there and going back again... It doesn't feel modern. How do you think about this. I'm spawn to spawn a feature branch on this branch to see the effects right away.

ewuerger:
Came back to see that I forget to tell you my whole plan I have in mind. I'm looking forward in discussing this with you. Seeing the code blocks in the great documentation you wrote I noticed that we don't have a copy-button. That feels off. So we'll fix that with [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io/en/latest/). Don't rustle your jimmies yet: The reason why I annoyed you so hard on cross-referencing, is that I want to have [sphinx-hoverxef](https://sphinx-hoverxref.readthedocs.io/en/latest/) such that you don't need to click, wait (download in germany yay) and after comprehending why we send you there and going back again... It doesn't feel modern. How do you think about this. I'm about to spawn a feature branch on this branch to see the effects right away.

docs/source/conf.py
```diff
@@ -44,6 +44,7 @@
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.coverage",
+    "sphinx.ext.napoleon",
```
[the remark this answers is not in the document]
Wuestengecko: Let's move the more general documentation- and Sphinx-related things to a separate discussion / PR, and keep this one focused on the declarative modelling stuff. That's most likely something that the other team members would also be interested in anyways, and it'd be good to hear their ideas as well. :)

docs/source/start/declarative.rst
```diff
@@ -0,0 +1,134 @@
...
+
+   my_model.save()
+
+Format description
+==================
+
+The expected YAML follows a simple format, where a parent object (i.e. an
+object that already exists in the model) is selected, and one or more of three
+different operations is applied to it: ``create``-ing new child objects,
+``modify``-ing the object itself, or ``delete``-ing one or more children.
```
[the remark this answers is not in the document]
Wuestengecko: > Can you explain why you made `_OPERATIONS` and its values `_operate_create`, `_operate_modify` and `_operate_delete` private?

They're private because the very fact that they exist is an implementation detail – they're not supposed to be called directly, and therefore have no reason to show up in the documentation of the public API.

If there's interesting documentation about the behavior of private functions/methods, it should be placed on the public functions that call them, i.e. in this case `apply`.

docs/source/start/declarative.rst
```diff
@@ -0,0 +1,134 @@
...
+Since v0.5, py-capellambse supports declarative modelling with the
+``capellambse.decl`` module.
+
+The YAML-based declarative modelling engine combines a few simple concepts into
+a powerful, but easy to use file format. These files can then be applied to any
+model supported by py-capellambse.
+
+Example
+=======
+
```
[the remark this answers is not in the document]
Wuestengecko: Good idea. Although I've moved the CLI heading down a bit until after the example, because that is common to both options.

docs/source/start/declarative.rst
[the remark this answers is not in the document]
Wuestengecko: > What do you think about converting this enumeration of available operations to an unordered list:

Another great idea! Added that as well. :)

Wuestengecko:
> I noticed that we don't have a copy-button. That feels off. So we'll fix that with [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io/en/latest/). […] I'm about to spawn a feature branch on this branch to see the effects right away.

I think that's a nice feature to have, but I also think that this branch is the wrong place to test it out. ;)

That is to say that of course you can try it out locally using this branch, but I won't be accepting any general documentation-related PRs into this one to keep it focused on declarative modelling. We can talk about general improvements to the docs as a whole in a separate PR / discussion.

docs/source/index.rst
```diff
@@ -42,6 +46,7 @@ tutorial section.
    start/intro-to-api
    start/how-to-explore-capella-mm
    start/developing-docs
+   start/declarative
```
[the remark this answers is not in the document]
Wuestengecko: Hm, I haven't really thought about where best to put it, I just added it at the end.

I think it would definitely make sense to keep the metamodel exploration and docs development together, because they're both targeted at contributors to the library, whereas the new page targets users. So the first one would make the most sense, where it's right after the introduction into the general API.

capellambse/decl.py
Wuestengecko: > promises would be pushed back into it

Not quite. I've been there during development as well, and it ended in an infinite loop, because it would take the first instruction out, see that it can't resolve the promise, and put it back ... into an empty queue, so that the next instruction is the exact same one it just took out. :clown_face: So we need at least two queues, one to park promises that can't be resolved yet. (This is by the way exactly what `tests/test_decl.py::TestApplyPromises::test_unfulfilled_promises_raise_an_exception` is for.)

Wuestengecko: > I suggest we take a closer look into special cases like re-locating `ModelObject`s and how this changes execution order of instructions.

It shouldn't affect the order of execution at all, because these are totally unrelated. But it could absolutely change the result of queries like `parent: la.root_function.functions[4]` (if those ever get implemented). We might even want to go as far as to resolve all `parent:` queries ahead of time, before modifying anything, in order to guarantee consistent results. But that's a discussion for another time :)

Wuestengecko: > the effects of moving elements from top to bottom (in XML) and vice versa

The only effect that it should have (assuming no bugs) is for the moved element to show up somewhere else. This would of course affect all lists that would have found it and shift the indexes of elements after it.

I can see a potential use case for moving elements into a specific position in such a list, instead of plainly appending it at the end. Do you think that should be explored further, or would it not be necessary?

capellambse/decl.py
Wuestengecko: Looking closely, this bit was actually missed by the unit tests.

I've added one, and can now confidently answer:

> Does this delattr call modify the XML code?

Yes, it does! :)

> If so how?

[The same way getattr/setattr-like statements do.](https://docs.python.org/3/reference/datamodel.html#implementing-descriptors)

capellambse/decl.py
Wuestengecko: Addressed by adding both - proper documentation *and* the missing dependency. :)

docs/source/index.rst
```diff
@@ -29,6 +29,10 @@ using Polarsys' Capella_ with Python. Common usage for this API:
 Additionally and as a core idea it provides an interface for the underlying
 database of the Capella model.
 
+Since v0.5, it also supports a simple, but powerful :ref:`declarative modelling
+language <declarative-modelling>`, which is based on the API for the semantic
+model.
+
```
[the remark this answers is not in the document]
ewuerger: I'm feeling weird. Since [there is no option to review unchanged files](https://github.com/community/community/discussions/9099) I add my code suggestion for `conf.py` in this comment.

```
extensions = [
    "sphinx.ext.autodoc",
    "sphinx.ext.coverage",
    "sphinx.ext.extlinks",
    "sphinx_rtd_theme",
    "jinja_in_rst",
    "capellambse.sphinx",
    "sphinx.ext.autosectionlabel",
]
# ...
# -- Options for extlinks ----------------------------------------------------
extlinks = {
    "version": (
        "https://github.com/DSD-DBS/py-capellambse/releases/tag/%s",
        "%s",
    )
}
```

docs/source/start/declarative.rst
```diff
@@ -0,0 +1,148 @@
...
+Format description
+==================
+
+The expected YAML follows a simple format, where a parent object (i.e. an
+object that already exists in the model) is selected, and one or more of three
+different operations is applied to it:
+
+- ``create``-ing new child objects,
+- ``modify``-ing the object itself, or
+- ``delete``-ing one or more children.
```
ewuerger, line 74: What do you think about documenting the modify and delete operation also? IMO `create` is covered very well and indepth, but we leave room for ✨creativity✨ of users... and as you know users can be very ✨creative✨.

In a future PR we could add other operations here like `extend`, etc.

ewuerger (changes requested):
Only documentation of `modify` and `delete` operations that comes close to documentation of `create` keeps us from 
                                                                                             ✨LGTM✨.

docs/source/start/declarative.rst
Wuestengecko: Good point. Not sure how I managed to miss that. I'm gonna write up something more about it.

capellambse/decl.py
Wuestengecko: As far as I'm aware, there was no interface spec for this use case yet. So I'd tend to agree with you and defer it to when someone (likely us) actually needs it, and we know exactly what the requirements are.

docs/source/index.rst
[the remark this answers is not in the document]
ewuerger: Just ignore this for once. In future PRs I'll stash these and spawn a completely new PR after the merge.

ewuerger (approved):
LGTM

ewuerger (approved):
LGTM

PR #147, opened 2022-09-16
Minor bug fixes

ewuerger:
- Fixed type annotation of `EnumAttributeProperty` and fixed the return value: Return the name instead of the value of the enum.
- Fixed a faulty attribute name on `PhysicalFunctionPkg`: components > functions and passed the correct class for the rootfunction: `PhysicalComponent` > `PhysicalFunction`.

### DIFF
--- a/capellambse/loader/xmltools.py
+++ b/capellambse/loader/xmltools.py
@@ -429,7 +429,7 @@ class EnumAttributeProperty(AttributeProperty):
             return None
         return self.enumcls[rawvalue]
 
-    def __set__(self, obj: t.Any, value: t.Any) -> None:
+    def __set__(self, obj: t.Any, value: str | enum.Enum) -> None:
         assert self.__objclass__ is not None
         if isinstance(value, str):
             try:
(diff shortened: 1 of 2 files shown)


Wuestengecko (commented):
Hmm, these exact two commits are apparently already in the history of `master`, with the exact same messages except for formatting:

```
❯ git range-diff b690a08 3f93e32 6177926
1:  8448952 ! 1:  69f760e fix: EnumAttributeProperty's `__set__`
    @@ Metadata
     Author: ewuerger <ernst.wuerger@gmail.com>

      ## Commit message ##
    -    fix: EnumAttributeProperty's `__set__`
    +    fix: EnumAttributeProperty's __set__

         Call `super.__set__` with `value.name` instead of the int value.

2:  3f93e32 = 2:  6177926 fix(layers): Fix passed Class and rename attribute
❯ git merge-base --is-ancestor 3f93e32 origin/master && echo yes
yes
```

I'll try to figure out what's going on there first before merging.

Wuestengecko:
I appreciate the effort, but yes, like I said these two commits are already in `master`. So no need to merge them again. :)

PR #155, opened 2022-10-06
Add 6.0 support

ewuerger:
Capella 6.0 doesn't introduce mayor changes to the metamodel such that no changes to our model submodule are needed. Only the `check_plugin_version` function is complaining since we fixated maximum supported versions of all plugins in the NAMESPACES global dictionary. This is a copy of the namespaces defined on the root element found in the `.capella` (or `.melodymodeller`) file. We figured that only constructing a `MelodyModel` instance for 1.3 was tested which isn't enough to claim full support of this legacy Capella version. Almost all model elements are accessed lazily. We dropped the test model 1.3/1.4 and migrated the parser and PVMT test models to 5.0 to claim a standardized baseline of supported **Capella version 5.0**.

Fix plugin version check:
Check if the plugin version is in the range of minimum and maximum supported versions (`["5.0.0", "6.0.0"]`)

This happens during `xtype_of` calls but only on elements that don't have an xtype... which should only happen on the root element in the `*.capella` file.


Wuestengecko:
af514f3 still has a "WIP" prefix, can you please double-check that you pushed the right thing before I spend the time to review it, and remove that prefix?

ewuerger:
> [af514f3](https://github.com/DSD-DBS/py-capellambse/commit/af514f3ef8be6d20a62a47b6d614a32eeffb595b) still has a "WIP" prefix, can you please double-check that you pushed the right thing before I spend the time to review it, and remove that prefix?

Reworked the history and added missing tests.

capellambse/helpers.py
```diff
@@ -558,7 +558,10 @@ def xtype_of(
 
     plugin_name, plugin_version = symbolic_ns[0][0], symbolic_ns[0][1]
     if not capellambse.check_plugin_version(plugin_name, plugin_version):
-        raise ValueError(f"Not handled version {ns!r}")
+        raise capellambse.UnsupportedVersionError(
+            f"Plugin '{plugin_name}' with unsupported version: "
+            f"'{plugin_version.version}'"
+        )
```
Wuestengecko, line 564: Since we already have a custom `UnsupportedVersionError` class, we could also make it a little easier to programmatically handle the error. We could raise it e.g. with:

```python
UnsupportedVersionError(plugin_name, actual_version, supported_min_version, supported_max_version)
```

and then override the `__str__` method in the exception class to transform it into a human readable text like above.

Wuestengecko (changes requested):

tests/test_model_loading.py
```diff
@@ -218,28 +218,59 @@ def test_http_file_handlers_passed_through_custom_headers(
...
 
 @pytest.fixture
-def model_db_path_with_patched_version(
-    request, tmp_path: pathlib.Path
+def model_path_with_patched_version(
+    request: pytest.FixtureRequest, tmp_path: pathlib.Path
 ) -> pathlib.Path:
-    """Indirect model database path fixture."""
-    tmp_dest = tmp_path / TEST_MODEL_5_0.parent.name
+    """Indirect parametrized fixture for version patched model database.
```
Wuestengecko, line 224:
```suggestion
    """Indirect parametrized fixture for version patched model.
```

capellambse/_namespaces.py
```diff
@@ -3,28 +3,63 @@
...
+    """Raised when :attr:`plugin` is unknown."""
+
+    plugin: Plugin
+
+    def __init__(self, plugin: Plugin, *args: object) -> None:
+        super().__init__(*args)
+        self.plugin = plugin
+
+    def __str__(self) -> str:
+        return f"Unknown plugin '{self.plugin!r}' is not supported."
```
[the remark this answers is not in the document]
ewuerger: Was changed as suggested by one of your previous [comments](https://github.com/DSD-DBS/py-capellambse/pull/155#discussion_r990013741).

capellambse/_namespaces.py
```diff
@@ -3,28 +3,63 @@
...
+
+
+class UnsupportedPluginVersionError(UnsupportedPluginError):
```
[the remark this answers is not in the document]
ewuerger: I prefer to keep both exceptions since the cause of the exception is simply clear just from the name of the exception. Both will inherit from the `ValueError`.

tests/test_model_loading.py
```diff
@@ -218,28 +218,59 @@ def test_http_file_handlers_passed_through_custom_headers(
...
-    )
-    model_db_file.write_text(patched_db, encoding="utf-8")
-    return model_db_file
+    request_param: str | tuple[str, str] = request.param  # type: ignore
+    for suffix in (".aird", ".capella"):
+        model_file = (tmp_dest / TEST_MODEL).with_suffix(suffix)
+        if isinstance(request_param, tuple):
+            re_params = request_param
+        else:
+            re_params = ("5.0.0", request_param)
```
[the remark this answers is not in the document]
ewuerger: That [requires some shenaningans with an `XSL` file](https://stackoverflow.com/a/51660868/5628391) and I didn't want to invest time in that just for an indirect fixture, used by test-cases for plugin version checking that we might drop in the future again(?). Altering the namespace with LXML is not just changing the `.nsmap` dict. From the [docs](https://lxml.de/api/lxml.etree._Element-class.html#nsmap):
image

capellambse/_namespaces.py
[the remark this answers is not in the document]
Wuestengecko: Looks good. I would change the text from "does not match" to "is outside of", but that's only another minor thing.

capellambse/_namespaces.py
```diff
@@ -132,9 +112,9 @@ def get_keys_and_plugins_from_namespaces_by_url(
     ------
     UnsupportedPluginError
         If requested ``url`` is not found in :class:`NAMESPACES`.
-    ValueError
-        If a namespace key, plugin-name pair is found more than once
-        in :class:`NAMESPACES`.
+    AssertionError
+        If a plugin is registered multiple times
+        (with different namespace-keys) in :class:`NAMESPACES`.
```
Wuestengecko, line 117: `assert` statements shouldn't be documented. In fact they mustn't, because they will get stripped by `python -O`.

capellambse/aird/parser/__init__.py
```diff
@@ -20,15 +20,16 @@
 from lxml import etree
 
 import capellambse
+from capellambse import _namespaces as _n
```
Wuestengecko, line 23: In other places we use `import collections.abc as cabc`, it would be good to unify this to use the same format:

```suggestion
import capellambse._namespaces as _n
```

capellambse/aird/parser/_common.py
```diff
@@ -17,13 +17,14 @@
 from lxml import builder, etree
 
 import capellambse.loader
-from capellambse import NAMESPACES, aird, helpers
+from capellambse import _namespaces as _n
```
Wuestengecko, line 20: Same as above:

```suggestion
import capellambse._namespaces as _n
```

capellambse/helpers.py
```diff
@@ -23,8 +23,9 @@
 from PIL import ImageFont
 
 import capellambse
+from capellambse import _namespaces as _n
```
Wuestengecko, line 26: Same as above:

```suggestion
import capellambse._namespaces as _n
```

capellambse/loader/core.py
```diff
@@ -27,6 +27,7 @@
 from lxml import etree
 
 import capellambse
+from capellambse import _namespaces as _n
```
Wuestengecko, line 30: Same as above:

```suggestion
import capellambse._namespaces as _n
```

capellambse/pvmt/model.py
```diff
@@ -5,8 +5,7 @@
 
 from __future__ import annotations
 
-from capellambse import NAMESPACES
-
+from .. import _namespaces as _n
```
Wuestengecko, line 8: Same as above:

```suggestion
import capellambse._namespaces as _n
```

capellambse/pvmt/types.py
```diff
@@ -8,21 +8,21 @@
 import abc
 import collections.abc as cabc
 
-from capellambse import NAMESPACES
+from capellambse import _namespaces as _n
```
Wuestengecko, line 11: Same as above:

```suggestion
import capellambse._namespaces as _n
```

capellambse/pvmt/validation.py
```diff
@@ -7,9 +7,9 @@
 import operator
 import re
 
-import capellambse
-from capellambse import get_keys_and_plugins_from_namespaces_by_url, helpers
+from capellambse import helpers
 
+from .. import _namespaces as _n
```
Wuestengecko, line 12: Same as above:

```suggestion
import capellambse._namespaces as _n
```

capellambse/_namespaces.py
```diff
@@ -181,41 +162,16 @@ def check_plugin(name: str, plugin: Plugin) -> None:
             plugin version range from :class:``capellambse.NAMESPACES``.
     """
     try:
-        my_plugin = NAMESPACES.get_plugin(name)
+        my_plugin = NAMESPACES_PLUGINS[name]
     except KeyError as err:
-        raise UnsupportedPluginError(plugin) from err
+        raise UnsupportedPluginError(f"{plugin}") from err
```
Wuestengecko, line 167: This looks weird. Wouldn't `str(plugin)` be cleaner (and possibly even faster)?

```suggestion
        raise UnsupportedPluginError(str(plugin)) from err
```

capellambse/helpers.py
```diff
@@ -23,8 +23,9 @@
...
 
-ATT_XT = f'{{{capellambse.NAMESPACES["xsi"]}}}type'
+ATT_XT = f'{{{_n.NAMESPACES["xsi"]}}}type'
```
Wuestengecko, line 28:
```suggestion
ATT_XT = f"{{{_n.NAMESPACES['xsi']}}}type"
```

capellambse/pvmt/types.py
```diff
@@ -8,21 +8,21 @@
...
+XTYPE_KEY = f"{{{_n.NAMESPACES['xsi']}}}type"
 
 
 class GenericPropertyValue(Generic, metaclass=abc.ABCMeta):
     """Base class for property value types."""
 
     description = AttributeProperty("xml_element", "description")
     xtype = AttributeProperty(
-        "xml_element", f'{{{NAMESPACES["xsi"]}}}type', writable=False
+        "xml_element", f'{{{_n.NAMESPACES["xsi"]}}}type', writable=False
```
Wuestengecko, line 25:
```suggestion
        "xml_element", f"{{{_n.NAMESPACES['xsi']}}}type", writable=False
```

capellambse/pvmt/types.py
```diff
@@ -45,7 +45,7 @@ def applyto(pvmt_ext, defelem, modelobj, targetelem):
         targetelem
             The new ``ownedPropertyValues`` element.
         """
-        for attr in (f'{{{NAMESPACES["xsi"]}}}type', "name", "value"):
+        for attr in (f'{{{_n.NAMESPACES["xsi"]}}}type', "name", "value"):
```
Wuestengecko, line 48:
```suggestion
        for attr in (f"{{{_n.NAMESPACES['xsi']}}}type", "name", "value"):
```

capellambse/pvmt/types.py
```diff
@@ -350,7 +350,7 @@ def applyto(cls, pvmt_ext, xml_element, groupname):
         groupelem = xml_element.makeelement(
             "ownedPropertyValueGroups",
             attrib={
-                f'{{{NAMESPACES["xsi"]}}}type': (
+                f'{{{_n.NAMESPACES["xsi"]}}}type': (
```
Wuestengecko, line 353:
```suggestion
                f"{{{_n.NAMESPACES['xsi']}}}type": (
```

capellambse/pvmt/types.py
```diff
@@ -369,7 +369,7 @@ def applyto(cls, pvmt_ext, xml_element, groupname):
         ):
             propelem = groupelem.makeelement("ownedPropertyValues")
             PROPERTY_LOADER[
-                defelem.get(f'{{{NAMESPACES["xsi"]}}}type').split(":")[-1]
+                defelem.get(f'{{{_n.NAMESPACES["xsi"]}}}type').split(":")[-1]
```
Wuestengecko, line 372:
```suggestion
                defelem.get(f"{{{_n.NAMESPACES['xsi']}}}type").split(":")[-1]
```

capellambse/pvmt/validation.py
```diff
@@ -88,13 +85,11 @@ def _validate_group_scope_class(pvmt_ext, scopedesc, xml_element):
...
-        yield_key_and_version_from_namespaces_by_plugin(match.group(1))
-    )
-    assert version <= capellambse.NAMESPACES.get_version(nskey)
-
+    ns = match.group(1)
+    nskey, plugin = _n.get_keys_and_plugins_from_namespaces_by_url(ns)
+    _n.check_plugin(nskey, plugin)
     return (
-        xml_element.get(f'{{{capellambse.NAMESPACES["xsi"]}}}type', "")
+        xml_element.get(f'{{{_n.NAMESPACES["xsi"]}}}type', "")
```
Wuestengecko, line 92:
```suggestion
        xml_element.get(f"{{{_n.NAMESPACES['xsi']}}}type", "")
```

Wuestengecko (changes requested):

Wuestengecko (dismissed):

PR #157, opened 2022-10-06
Add a note for save and update_cache dependance

ewuerger:
If the `MelodyModel` was initialzed from a remote URL and the user intents to use `.save(push=True)` (i.e. `write_transaction` is executed) an up2date cache is needed. This is ensured by the default `update_cache=True` parameter. I felt the urge to update the docs with additional notes about this dependence at the right places.

Take notice that I used a `~` in `:class:`~capellambse.loader.filehandler.FileHandler`` to shorten the rendered reference to only `FileHandler`.

### DIFF
--- a/capellambse/loader/core.py
+++ b/capellambse/loader/core.py
@@ -382,6 +382,11 @@ class MelodyLoader:
     def save(self, **kw: t.Any) -> None:
         """Save all model files back to their original locations.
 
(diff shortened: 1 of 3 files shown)


Wuestengecko (approved):
Good idea. That's probably something most users don't have on their radar.

PR #158, opened 2022-10-07
Update docs for `diagram_cache=` arg of `MelodyModel`

Wuestengecko:


jamilraichouni:
We may also want to add that it must be ensure that the revision matches the default branch for the model's repo.

jamilraichouni (approved):
LGTM

PR #160, opened 2022-10-14
Documentation improvements

Wuestengecko:
This PR:

- [x] Enables the pydocstyle pre-commit hook
- [x] Enables and configures `docformatter`
- [x] Fixes all the issues with docstrings discovered by them, except for D1 issues (i.e. entirely missing docstrings). Most of the changes were reflowing paragraphs, adding missing punctuation, and adding missing documentation on function parameters.

~~@ewuerger Do we want to put the Sphinx theme change in here as well? I'd keep it as draft until then, otherwise it's ready.~~\
That's a little bit more than I anticipated, let's split that up so we can properly use Github's review tools.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,7 +27,7 @@ jobs:
           python -m pip install '.[docs]'
       - name: Auto-generate APIDOC sources
         run: |-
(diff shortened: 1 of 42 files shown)


ewuerger (approved):
LGTM

PR #162, opened 2022-10-18
Rename the `exchanges` attribute to `related_exchanges` on OperationalActivity

ewuerger:
In order to standardize the `.related_exchanges` attribute I propose to rename the `exchanges` property of `OperationalActivity` to `related_exchanges`. This makes my life easier in the context diagrams collection stage.

### DIFF
--- a/capellambse/model/layers/oa.py
+++ b/capellambse/model/layers/oa.py
@@ -38,6 +38,9 @@ class OperationalActivity(fa.AbstractFunction):
         operator.attrgetter("_model.oa.all_entities"),
         matchtransform=operator.attrgetter("activities"),
     )
+    exchanges = c.DirectProxyAccessor(
+        fa.FunctionalExchange, aslist=c.ElementList
+    )
 
     @property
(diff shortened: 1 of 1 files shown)


capellambse/model/layers/oa.py
```diff
@@ -48,7 +48,7 @@ def outputs(self) -> c.ElementList[fa.FunctionalExchange]:
         return self._model.oa.all_activity_exchanges.by_source(self)
 
     @property
-    def exchanges(self) -> c.ElementList[fa.FunctionalExchange]:
+    def related_exchanges(self) -> c.ElementList[fa.FunctionalExchange]:
```
Wuestengecko, line 51: That looks like an oversight from 59fc83257efcf78926f95234e6adebba81dc820c. IMO it should therefore be declared as "fix" rather than "refactor". Can you also add the "owned exchanges" as the new `exchanges` attribute? Should be a simple `DirectProxyAccessor` if I'm not entirely wrong.

PR #163, opened 2022-10-18
Fix class diagrams ([CDB])

ewuerger:
This PR improves the rendering of class diagrams (i.e. diagrams with class type **[CDB]**) provided by the svg submodule.
The following happened:

- Added 2 new filters:
    - `hide.role.names.filter`
    - `hide.association.labels.filter`
- The Association edge factory was fixed. There was a bug that caused wrong labels (only navigable role names) to appear in the SVG. The factory now does only patch the style class from the `aggregationKind` attribute of the first melodyobject. From this syleclass (Association (per default), Aggregation and Composition) we infer which end-marker is to be taken and added to the defs section of the drawing.
- The `FilledDiamondMarker` for the `Aggregation` styleclass was added to decorations.

The bugs around bendpoints calculation causing unorthogonal routing still prevails. I was even able to run into a `Lines are parallel` exception in the `line_intersect` method. Full of joy I await the refactoring of our internal aird-parser and svg calculations...

### DIFF
--- a/capellambse/aird/parser/_filters/global.py
+++ b/capellambse/aird/parser/_filters/global.py
@@ -12,6 +12,22 @@ from capellambse import aird, helpers, model
 from . import FilterArguments, composite, global_filter
 
 XT_CEX_FEX_ALLOCATION = "org.polarsys.capella.core.data.fa:ComponentExchangeFunctionalExchangeAllocation"
+EXCHANGES_WITH_ROLES = frozenset(
+    ("Association", "Aggregation", "Composition", "Generalization")
+)
+
+
+@global_filter("hide.role.names.filter")
+def hide_role_names(args: FilterArguments, flt: lxml.etree._Element) -> None:
+    """Hide names of roles on an exchange."""
+    del flt
+    classes = EXCHANGES_WITH_ROLES
+    for obj in args.target_diagram:
+        if not isinstance(obj, aird.Edge):
+            continue
+
+        if len(obj.labels) > 1 and obj.styleclass in classes:
+            obj.labels = obj.labels[:1]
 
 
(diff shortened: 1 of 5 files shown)


capellambse/aird/parser/_filters/global.py
```diff
@@ -12,6 +12,40 @@
 from . import FilterArguments, composite, global_filter
 
 XT_CEX_FEX_ALLOCATION = "org.polarsys.capella.core.data.fa:ComponentExchangeFunctionalExchangeAllocation"
+EXCHANGES_WITH_ROLES = frozenset(
+    ("Association", "Aggregation", "Composition", "Generalization")
```
Wuestengecko, line 16:
```suggestion
    {"Aggregation", "Association", "Composition", "Generalization"}
```

(This matches the `repr` of frozensets)

capellambse/aird/parser/_filters/global.py
```diff
@@ -12,6 +12,40 @@
...
+) -> None:
+    """Hide names of roles on an exchange."""
+    del flt
+
+    classes = EXCHANGES_WITH_ROLES
+    for obj in args.target_diagram:
+        if not (isinstance(obj, aird.Edge) or obj.styleclass in classes):
+            continue
+
+        assert isinstance(obj, aird.Edge)
```
Wuestengecko, line 32: This would look more elegant (and wouldn't need a double-check in the form of an assertion) if we simply ignored everything that's not an Edge, i.e.:

```suggestion
        if not isinstance(obj, aird.Edge) or obj.styleclass not in classes:
            continue
```

Wuestengecko (changes requested):

Wuestengecko (approved):

Issue #159, opened 2022-10-10
Only accept *.aird file as path or entrypoint parameter for MelodyModel and MelodyLoader


ewuerger:
Looks like there's a bug that `MelodyModel` and `MelodyLoader` accept the path to a `.capella` file as `path` or `entrypoint` argument, even though it's missing important information.

_Originally posted by @Wuestengecko in https://github.com/DSD-DBS/py-capellambse/pull/155#discussion_r989988012_

PR #164, opened 2022-10-19
Disallow model loading from capella files without valid entrypoint

ewuerger:
Valid entrypoint file-types are `.aird` and `.airdfragment` coming from `VISUAL_EXT`. This solves #159.

Another solution would be to not raise an `InvalidEntrypoint` exception but rather look for valid entrypoints in the same directory of the given entrypoint. Assuming that modelfiles are kept together in the same directory. But this doesn't incentivise to pass the correct entrypoint which is the `.aird`/`.airdfragment` file.

### DIFF
--- a/capellambse/loader/core.py
+++ b/capellambse/loader/core.py
@@ -320,6 +320,8 @@ class MelodyLoader:
             else:
                 self.resources[resname] = reshdl
         self.entrypoint = self.__derive_entrypoint(entrypoint)
+        if self.entrypoint.suffix != ".aird":
+            raise ValueError("Invalid entrypoint, specify the ``.aird`` file")
 
         self.trees: dict[pathlib.PurePosixPath, ModelFile] = {}
         self.__load_referenced_files(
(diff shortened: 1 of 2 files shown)


Wuestengecko (changes requested):
I appreciate the effort, but this is a waaay overcomplicated way of spelling `if self.entrypoint.suffix != ".aird": raise ...`.

Also, `airdfragment` files are just as invalid as everything else that's not the main `.aird` file. The test case that tried to load one was wrong.

Wuestengecko (approved):

jamilraichouni:
> I appreciate the effort, but this is a waaay overcomplicated way of spelling `if self.entrypoint.suffix != ".aird": raise ...`.
> 
> Also, `airdfragment` files are just as invalid as everything else that's not the main `.aird` file. The test case that tried to load one was wrong.

`if self.entrypoint.suffix.lower() != ".aird": raise ...`

for Windows users and the theoretically possible case that someone has an .AIRD or ,Aird etc. file :-)

Wuestengecko:
>     if self.entrypoint.suffix.lower() != ".aird": raise ...

Interesting point @jamilraichouni, but ahkschually no (not at this point in time at least): We're doing case-sensitive comparisons all over the place during model loading anyways, using `lower()` here would only get it rejected at some later point. Making all of that case-insensitive as well would require a bit more work. And so far nobody complained, so ... ¯\\\_(ツ)\_/¯

PR #169, opened 2022-10-24
Fix physical links

ewuerger:
This PR enables context collection for `PhysicalComponent`s in the [context-diagrams-extension](https://github.com/DSD-DBS/capellambse-context-diagrams/pull/23).

### DIFF
--- a/capellambse/decl.py
+++ b/capellambse/decl.py
@@ -207,7 +207,16 @@ def _operate_modify(
             except _UnresolvablePromise as p:
                 yield p.args[0], {"parent": parent, "modify": {attr: value}}
                 continue
-        setattr(parent, attr, value)
+
+        if isinstance(value, list):
+            delattr(parent, attr)
+            yield from _create_complex_objects(promises, parent, attr, value)
+        elif isinstance(value, dict):
+            obj = getattr(parent, attr)
+            for k, v in value.items():
+                setattr(obj, k, v)
+        else:
+            setattr(parent, attr, value)
 
 
 def _resolve(
@@ -268,7 +277,9 @@ def _create_complex_objects(
             f" {type(parent).__name__}.{attr} is not model-coupled"
         )
     for child in objs:
-        if isinstance(child, (list, Promise, UUIDReference)):
+        if isinstance(
+            child, (common.GenericElement, list, Promise, UUIDReference)
+        ):
             try:
                 obj = _resolve(promises, parent, child)
             except _UnresolvablePromise as p:
--- a/capellambse/model/common/accessors.py
+++ b/capellambse/model/common/accessors.py
@@ -120,6 +120,9 @@ class DeprecatedAccessor(Accessor[T]):
 
     def __get__(self, obj, objtype=None):
         self.__warn()
+        if obj is None:
+            return self
+
         return getattr(obj, self.alternative)
 
     def __set__(self, obj: element.GenericElement, value: t.Any) -> None:
--- a/capellambse/model/common/accessors.py
+++ b/capellambse/model/common/accessors.py
@@ -12,6 +12,8 @@ __all__ = [
     "DeepProxyAccessor",
     "LinkAccessor",
     "AttrProxyAccessor",
+    "PhysicalLinkEndsAccessor",
+    "IndexAccessor",
     "AlternateAccessor",
     "ParentAccessor",
     "CustomAccessor",
@@ -848,6 +850,67 @@ class AttrProxyAccessor(WritableAccessor[T], PhysicalAccessor[T]):
         obj._element.set(self.attr, " ".join(parts))
 
 
+class PhysicalLinkEndsAccessor(AttrProxyAccessor[T]):
+    def __init__(
+        self,
+        class_: type[T],
+        attr: str,
+        *,
+        aslist: type[element.ElementList],
+    ) -> None:
+        super().__init__(class_, attr, aslist=aslist)
+        assert self.aslist is not None
+        self.aslist.fixed_length = 2
+
+
+class IndexAccessor(Accessor[T]):
+    """Access a specific index in an ElementList of a fixed size."""
+
+    __slots__ = ("index", "wrapped")
+
+    def __init__(self, wrapped: str, index: int) -> None:
+        super().__init__()
+        self.index = index
+        self.wrapped = wrapped
+
+    @t.overload
+    def __get__(self, obj: None, objtype=None) -> Accessor:
+        ...
+
+    @t.overload
+    def __get__(self, obj, objtype=None) -> T | element.ElementList[T]:
+        ...
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return self
+        container = getattr(obj, self.wrapped)
+        if not isinstance(container, element.ElementList):
+            raise RuntimeError(
+                f"Cannot get {self._qualname}: {self.wrapped} is not a list"
+            )
+        if len(container) <= self.index:
+            raise RuntimeError(
+                f"Broken XML: Expected at least {self.index + 1} elements,"
+                f" found {len(container)}"
+            )
+        return container[self.index]
+
+    def __set__(self, obj: element.GenericElement, value: t.Any) -> None:
+        container = getattr(obj, self.wrapped)
+        if not isinstance(container, ElementListCouplingMixin):
+            raise TypeError(
+                f"Cannot set {self._qualname}:"
+                f" {self.wrapped} is not a coupled list"
+            )
+        if len(container) < self.index:
+            raise RuntimeError(
+                f"Broken XML: Expected at least {self.index+1} elements,"
+                f" found {len(container)}"
+            )
+        container[self.index] = value
+
+
 class AlternateAccessor(Accessor[T]):
     """Provides access to an "alternate" form of the object."""
 
@@ -1220,6 +1283,7 @@ class ElementListCouplingMixin(element.ElementList[T], t.Generic[T]):
     """
 
     _accessor: WritableAccessor[T]
+    fixed_length: t.ClassVar[int] = 0
 
     def __init__(
         self, *args: t.Any, parent: element.ModelObject, **kw: t.Any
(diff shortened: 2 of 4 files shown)


capellambse/model/crosslayer/cs.py
```diff
@@ -75,6 +76,23 @@ class PhysicalLink(PhysicalPort):
 
     physical_paths: c.Accessor
 
+    @property
+    def linkEnds(self) -> tuple[PhysicalPort, PhysicalPort]:
```
Wuestengecko, line 80: Wrong type annotation, this returns an `ElementList[PhysicalPort]`. Weird, mypy should complain about this ...

Wuestengecko (changes requested):

capellambse/model/crosslayer/cs.py
```diff
@@ -66,7 +67,7 @@ class PhysicalPort(c.GenericElement):
 class PhysicalLink(PhysicalPort):
     """A physical link."""
 
-    linkEnds = c.AttrProxyAccessor(
+    link_ends = c.AttrProxyAccessor(
```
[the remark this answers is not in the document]
ewuerger: I checked and confirmed that with the current Capella version (5.0, 5.2 and 6.0) the `PhysicalLink` can only have 2 elements (`PhysicalPort`s) in their `linkEnds` attribute. Imo I'd say that an individual `Accessor` is not necessary since I've encountered an link attribute with such an implicit length condition only once: Exactly here.

ewuerger:
Everything is already implemented in `__setattr__` of `GenericElement`. So we call this again in the newly added setters. Please have another look.

Wuestengecko (dismissed):

tests/test_xlayer_cs.py
```diff
@@ -1,10 +1,12 @@
 # SPDX-FileCopyrightText: Copyright DB Netz AG and the capellambse contributors
 # SPDX-License-Identifier: Apache-2.0
 
+import pytest
+
 from capellambse import MelodyModel
 
 
-def test_physical_path_has_ordered_list_of_involved_items(model: MelodyModel):
+def test_PhysicalPath_has_ordered_list_of_involved_items(model: MelodyModel):
```
ewuerger, line 9: Thank you for also making all names of the tests consistent. This keeps my sanity up high.

tests/test_xlayer_cs.py
```diff
@@ -31,19 +33,48 @@ def test_physical_path_has_ordered_list_of_involved_links(model: MelodyModel):
...
+    source_pp = model.by_uuid("76d9c301-c0ad-4615-9f02-b804b018decf")
+    target_pp = model.by_uuid("53c9fe29-18e2-4642-906e-b7507bf0ff39")
+
+    link.source = source_pp
+    link.target = target_pp
+
+    assert source_pp == link.ends[0]
+    assert source_pp == link.source
+    assert target_pp == link.ends[1]
+    assert target_pp == link.target
```
ewuerger, line 80: Nice and clean arrangement.

tests/test_xlayer_cs.py
```diff
@@ -31,19 +33,48 @@ def test_physical_path_has_ordered_list_of_involved_links(model: MelodyModel):
...
+    expected = [source_pp, target_pp]
+    assert getattr(link, attr_name, None) != expected
+
+    setattr(link, attr_name, [source_pp, target_pp])
+    actual = getattr(link, attr_name)
+
+    assert actual == expected
+    assert [link.source, link.target] == expected
+
+
```
ewuerger, line 68:
```suggestion
    old_ends = getattr(link, attr_name, None)

    setattr(link, attr_name, [source_pp, target_pp])
    actual = getattr(link, attr_name)

    assert old_ends != expected
    assert actual == expected
    assert [link.source, link.target] == expected

```
I like this more b/c of the prepare, act, assert pattern/convention.

capellambse/decl.py
```diff
@@ -197,7 +197,17 @@ def _operate_modify(
...
+        if isinstance(value, list):
+            delattr(parent, attr)
+            yield from _create_complex_objects(promises, parent, attr, value)
+        else:
+            if isinstance(value, dict):
+                obj = getattr(parent, attr)
+                for k, v in value.items():
+                    setattr(obj, k, v)
+            else:
+                setattr(parent, attr, value)
```
ewuerger, line 210:
```suggestion
        elif isinstance(value, dict):
            obj = getattr(parent, attr)
            for k, v in value.items():
                setattr(obj, k, v)
        else:
            setattr(parent, attr, value)
```

I like this a bit more with one level less of indentation. Semantically no change.

capellambse/model/common/accessors.py
```diff
@@ -610,15 +616,18 @@ def __get__(self, obj, objtype=None):
         return self._make_list(obj, elems)
 
     def __set__(self, obj, value):
-        if self.aslist is not None:
-            self.__get__(obj)[:] = value
-            return
-
+        if self.aslist is None:
+            if isinstance(value, cabc.Iterable):
+                raise TypeError(f"{self._qualname} expects a single element")
```
ewuerger, line 621: Doesn't this fire if value is a string (and `self.aslist is None`?

capellambse/model/common/accessors.py
```diff
@@ -610,15 +616,18 @@ def __get__(self, obj, objtype=None):
...
+            value = (value,)
+        elif not isinstance(value, cabc.Iterable):
+            raise TypeError(f"{self._qualname} expects an iterable")
```
ewuerger, line 624: Essentially same issue: When value is a string it would get through here.

capellambse/model/common/accessors.py
```diff
@@ -1201,16 +1272,23 @@ def __setitem__(
...
+        accessor = type(self)._accessor
+        assert isinstance(accessor, WritableAccessor)
+
+        new_objs = list(self)
+        new_objs[index] = value  # type: ignore[index]
+
+        if self.fixed_length and len(new_objs) != self.fixed_length:
+            raise TypeError(
+                f"Cannot set: List must stay at length {self.fixed_length}"
+            )
```
ewuerger, line 1284: Isn't a `ValueError` more suited for this? `value`s length doesn't match to this `ElementList.fixed_length`.

capellambse/model/common/accessors.py
```diff
@@ -1201,16 +1272,23 @@ def __setitem__(
...
+
+        accessor.__set__(self._parent, new_objs)
 
     def __delitem__(self, index: int | slice) -> None:
+        if self.fixed_length and len(self) <= self.fixed_length:
+            raise TypeError("Cannot delete from a fixed-length list")
```
ewuerger, line 1290: `TypeError` seems weird to me. The error condition is that this `ElementList` has a `fixed_length` and removing from it would cause a length violation: There's one less element in it opposing the constraint.

I'd expect either a custom exception or a `ValueError` here too with a different message. Deletion can be done if somehow there are more elements written in the XML via Capella and `len(self) >= self.fixed_length + 1`. Else why is there functional code afterwards?

capellambse/model/common/accessors.py
```diff
@@ -1250,6 +1328,9 @@ def create(self, *type_hints: str | None, **kw: t.Any) -> T:
             Initialize the properties of the new object. Depending on
             the object, some attributes may be required.
         """
+        if self.fixed_length and len(self) >= self.fixed_length:
+            raise TypeError("Cannot create elements in a fixed-length list")
```
ewuerger, line 1332: Here again. Imagine having an empty fixed-length `ElementList`. One can still populate up to `fixed_length`-many elements. It is just that this addition would exceed the `fixed_length` constraint.

capellambse/model/common/accessors.py
```diff
@@ -1270,6 +1351,9 @@ def create_singleattr(self, arg: t.Any) -> T:
             The method to override in Accessors in order to implement
             this operation.
         """
+        if self.fixed_length and len(self) >= self.fixed_length:
+            raise TypeError("Cannot create elements in a fixed-length list")
```
ewuerger, line 1355: Same as above here.

capellambse/model/common/accessors.py
```diff
@@ -1288,6 +1372,9 @@ def delete_all(self, **kw: t.Any) -> None:
             del self[index]
 
     def insert(self, index: int, value: T) -> None:
+        if self.fixed_length and len(self) >= self.fixed_length:
+            raise TypeError("Cannot insert into a fixed-length list")
```
ewuerger, line 1376: Same argumentation as above.

ewuerger (commented):
I can't request changes, only comment. But I want to. Please either resolve my issues/apply-suggestions or try to convince me otherwise.

tests/test_xlayer_cs.py
Wuestengecko: I'd rather not, for the same reason I outlined in [another comment](https://github.com/DSD-DBS/py-capellambse/pull/182#discussion_r1011508921).

capellambse/model/common/accessors.py
Wuestengecko: I'm not sure if a `str` is even valid as `value` here in the first place, but even if not, we should guard against it anyways. (This stuff is annoying. Thanks for looking out for it too.)

capellambse/model/common/accessors.py
Wuestengecko: The idea behind TypeError is that "this type of object doesn't allow this operation" (which I've seen used in this way in a few places). And it makes sense because it's a different type of "CoupledElementList" than one that would allow it (because `fixed_length` is a `ClassVar`). But "the value you provided doesn't fit" works just as well ... I don't know, I'm biased towards mine, so you call it. :P

capellambse/model/common/accessors.py
Wuestengecko: > Deletion can be done if somehow there are more elements written in the XML via Capella and `len(self) >= self.fixed_length + 1`. Else why is there functional code afterwards?

Items can also be deleted if the list isn't `fixed_length` to begin with, which would be the typical case to hit that code.

capellambse/model/common/accessors.py
ewuerger: Your explanation fits.

ewuerger:
LGTM.

PR #175, opened 2022-10-27
docs: Set type-hints format to `short`

ewuerger:
During a review I noticed that we have long hyperlinks for the type annotations that come from **autodoc** implemented in [PR#7018](https://github.com/sphinx-doc/sphinx/pull/7018) (not napoleon (!) ). Since we have a very structured code base with many submodules I want to shorten these. It is accomplished with the [`autodoc_typehints_format = "short"`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints_format) configuration. This should be the default configuration but somehow it is not. Do we need to upgrade sphinx?

The effects can be seen here:
**`autodoc_typehints_format = "fully-qualified"`**:
image

**`autodoc_typehints_format = "short"`**:
image

Unfortunately the typehints that use a mixture of `typing.*` or `collections.abc.*` and our own types are not shortened (example: **aslist** parameter). This can be fixed by introducing nice type aliases and maintain them in the `conf.py` via [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases). Or maybe even using [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) extension for this?

Do you think further investigation is worth it?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -74,6 +74,7 @@ rst_epilog = f"""
 autoclass_content = "class"
 autodoc_class_signature = "separated"
 autodoc_typehints = "description"
+autodoc_typehints_format = "short"
 
 
 # -- Options for napoleon ----------------------------------------------------


Wuestengecko:
> This should be the default configuration but somehow it is not. Do we need to upgrade sphinx?

According to the docs, it became the default with Sphinx 5.0, which we are using in our CI: https://github.com/DSD-DBS/py-capellambse/actions/runs/3339631666/jobs/5528814069

And the setting takes effect in both the GH pages and on my local dev machine. https://dsd-dbs.github.io/py-capellambse/code/capellambse.model.common.html

So I don't see the issue you're trying to solve? Do you perhaps just need to upgrade Sphinx locally on your machine?

ewuerger:
> So I don't see the issue you're trying to solve? Do you perhaps just need to upgrade Sphinx locally on your machine?

Yes, my sphinx version was <5.3.0. I close this.

PR #182, opened 2022-11-01
Add `extend` operation to Declarative modelling

ewuerger:
This key is used for solving the relocation/movement of existing(to-be-created) model objects.


tests/test_decl.py
```diff
@@ -451,6 +451,99 @@ def test_delete_operations_delete_attributes_if_no_list_of_uuids_was_given(
...
 
 
+class TestApplyExtend:
+    @staticmethod
+    def test_extend_operations_change_model_objects_in_place(
+        model: capellambse.MelodyModel,
+    ) -> None:
+        root_component = model.by_uuid(ROOT_COMPONENT)
+        fnc = model.by_uuid(ROOT_FUNCTION)
+        in_functions_before_apply = fnc in root_component.functions
```
Wuestengecko, line 461: This should be directly `assert`ed here, because if this precondition isn't met, it doesn't make sense to even begin with the test. Fail fast. (Consider "checking whether the test environment makes sense" to be part of "preparing the test environment".)

If you want to make it *really* explicit, add a message to the assertion stating as much, like:

```python
assert fnc not in root_component.functions, "Precondition failed"
```

tests/test_decl.py
```diff
@@ -451,6 +451,99 @@ def test_delete_operations_delete_attributes_if_no_list_of_uuids_was_given(
...
+        yml = f"""\
+            - parent: !uuid {ROOT_COMPONENT}
+              extend:
+                functions:
+                  - !uuid {fnc.uuid}
```
Wuestengecko, line 466: Same as in the doc: Because components cannot "own" functions, this doesn't move the `fnc` but only creates a new reference to it. The best way to ensure that something was actually moved is checking whether its `.parent` changed:

```python
old_parent = fnc.parent
decl.apply(...)
assert old_parent != fnc.parent  # AssertionError
```

Please add this to the test. :)

tests/test_decl.py
```diff
@@ -451,6 +451,99 @@ def test_delete_operations_delete_attributes_if_no_list_of_uuids_was_given(
...
+        yml = f"""\
+            - parent: !uuid {ROOT_FUNCTION}
+              create:
+                functions:
+                  - name: {fnc_name}
+                    promise_id: promised-fnc
+            - parent: !uuid {ROOT_COMPONENT}
+              extend:
+                functions:
+                  - !promise promised-fnc
```
Wuestengecko, line 491: Again, referenced, not moved.

docs/source/start/declarative.rst
```diff
@@ -184,6 +187,44 @@ e.g. numeric properties. This example changes the ``min_card`` property of an
...
+
+   - parent: !uuid f28ec0f8-f3b3-43a0-8af7-79f194b29a2d
+     create:
+       functions:
+         - name: The promised one
+           promise_id: promised-fnc
+   - parent: !uuid 0d2edb8f-fa34-4e73-89ec-fb9a63001440
+       extend:
+         functions:
+           - !promise promised-fnc
```
Wuestengecko, line 224: This bit of functionality (i.e. the combination `extend:`/move + `!promise`) doesn't really make sense to me. Why would you want to create the function somewhere, and then later *in the same file* move it somewhere else - instead of just creating it there in the first place? This can (and probably will at some point) create an issue in more complex scenarios with multiple references, where the actual order of modifications in the model isn't reliably known beforehand. (Which isn't the point of declarative modelling to begin with.)

Trying to catch this case while applying a YAML would probably introduce a fair bit of complexity, but the least we can do here is to not encourage it by making it one of our documented examples.

Wuestengecko (changes requested):

docs/source/start/declarative.rst
```diff
@@ -73,7 +73,10 @@ object that already exists in the model) is selected, and one or more of three
 different operations is applied to it:
 
 - ``create``-ing new child objects,
-- ``modify``-ing the object itself, or
+- ``modify``-ing the object itself,
+- ``extend``-ing an
+  :py:class:`~capellambse.model.common.accessors.ElementListCouplingMixin`-ish
+  attribute on the object, or
```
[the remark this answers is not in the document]
ewuerger: So I thought till then we can refer to the API reference. [ElementList](https://dsd-dbs.github.io/py-capellambse/code/capellambse.model.common.html#capellambse.model.common.element.ElementList) isn't telling much for altering the model.

docs/source/start/declarative.rst
```diff
@@ -184,6 +187,44 @@ e.g. numeric properties. This example changes the ``min_card`` property of an
...
+.. code-block:: yaml
+   :emphasize-lines: 5,9
+
+   - parent: !uuid f28ec0f8-f3b3-43a0-8af7-79f194b29a2d
+     create:
+       functions:
+         - name: The promised one
+           promise_id: promised-fnc
+   - parent: !uuid 0d2edb8f-fa34-4e73-89ec-fb9a63001440
+       extend:
```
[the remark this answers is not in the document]
ewuerger: Don't know what you mean. We use 2 spaces for indenting here. If it would be syntactically wrong wouldn't sphinx tell me during building?

docs/source/start/declarative.rst
```diff
@@ -184,6 +187,44 @@ e.g. numeric properties. This example changes the ``min_card`` property of an
        min_card: 0
        max_card: .inf
 
+Extending container attributes of objects
+-----------------------------------------
```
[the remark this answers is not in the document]
ewuerger: I didn't know that this behavior exists and I find it just a bit missleading/unintuitive. For me `create` means that a new object is created... but one could also argue that used in combination with !uuid/!promise only a new reference is created. To complete the logic of the latter, `delete` should also distinguish between only deleting the reference (href from attribute, or w/e) and really deleting it. That means wiping the model from the objects existence without leaving broken references behind. This would happen if the user calls `delete` on an attribute where the `ProxyAccessor`(parent/child relationship) is used.

So either we keep this behavior of `create`, add a subsection to inform about this feature in the docs and make sure on `delete` completeness described above. Or we take the verbose route and add `extend` (maybe better: `create-reference`/`delete-reference`)... where I'd need to also implement the `remove` (`delete-reference`) operation.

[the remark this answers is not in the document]
Wuestengecko: > I didn't know that this behavior exists and I find it just a bit misleading/unintuitive.

That's what I meant with my last paragraph - not quite what you'd call the "obvious way to do it". :) I agree we should at least mention that it's a thing in the docs somewhere.

> make sure on `delete` completeness

This part should rather go into the base library, as it's not really specific to decl's `delete` operation. If you delete something with just `del my_func.functions[0]`, you'd probably expect that no broken refs stay behind (which is also Capella's behavior, by the way). So, yes, we should implement it, but not here.

This might be a somewhat bigger undertaking though, because we'd need to make an index of who links where if we don't want to completely destroy runtime performance, and whoever deletes also needs to understand how much to delete (e.g. the LinkAccessor needs to delete the entire XML element, while the AttrProxyAccessor only removes one UUID from one attribute).

> Or we take the verbose route and add `extend` (maybe better: `create-reference`/`delete-reference`)... where I'd need to also implement the `remove` (`delete-reference`) operation.

Not entirely sure what you're aiming at. Do you want to split up "create" into "create-object" and "create-reference", and then add "extend" on top for moving stuff around? Not a big fan of that, it just makes things unnecessarily complex for the end user *and* the implementation.

Currently there's no easy and reliable way to distinguish between the cases anyway. Objects get resolved successively as they become available, and by the time we finally get to act on one, we don't know anymore whether the original YAML said `!uuid`, `!promise`, or had a dictionary. I'd say this approach isn't feasible without rewriting large parts of the logic.

docs/source/start/declarative.rst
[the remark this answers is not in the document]
Wuestengecko: Fair point. Let's keep it like this for now and open an issue about it.

docs/source/start/declarative.rst
ewuerger: You know more than me about the intricacies of the declarative-modelling implementation. But creating new references to promised (i.e. elements to be created) elements should work, no?

tests/test_decl.py
```diff
@@ -206,6 +206,87 @@ def test_decl_can_create_simple_objects_by_passing_plain_strings(
...
+            - parent: !uuid {ROOT_FUNCTION}
+              extend:
+                functions:
+                  - !uuid {fnc.uuid}
+            """
+
+        decl.apply(model, io.StringIO(yml))
+
+        assert fnc in root_function.functions
+        assert fnc.parent == root_function != old_parent
```
Wuestengecko, line 229: Instead of this roundabout check at the end (where I'm not even sure if the ordering of operands is correct..? The first two might be swapped), it would IMO be cleaner to add another precondition to explicitly `assert old_parent != root_function`. (Even though technically that and the `root_function != old_parent` part here are redundant anyways, but in this case I agree with "explicit is better than implicit".)

tests/test_decl.py
```diff
@@ -206,6 +206,87 @@ def test_decl_can_create_simple_objects_by_passing_plain_strings(
...
+              extend:
+                functions:
+                  - !promise promised-fnc
+            """
+
+        decl.apply(model, io.StringIO(yml))
+        fnc_from_lookup = model.la.all_functions.by_name(fnc_name)
+        fnc_from_comp_functions = root_component.functions.by_name(fnc_name)
+
+        assert fnc_from_lookup == fnc_from_comp_functions
```
Wuestengecko, line 258: While technically mostly correct (I think), this assertion looks quite weird for what it's intended to do.

We should rather obtain one (persistent) reference to the object-under-test (i.e. the `ROOT_COMPONENT`) before executing `decl.apply()`, check as precondition that it does not have a "The promised one" function, and then check as test result that it now has one.

Wuestengecko (changes requested):

docs/source/start/declarative.rst
```diff
@@ -72,7 +72,7 @@ The expected YAML follows a simple format, where a parent object (i.e. an
 object that already exists in the model) is selected, and one or more of three
 different operations is applied to it:
 
-- ``create``-ing new child objects,
+- ``extend``-ing the object on listish attributes,
```
Wuestengecko, line 75: Typo.

```suggestion
- ``extend``-ing the object on list attributes,
```

docs/source/start/declarative.rst
```diff
@@ -84,8 +84,17 @@ our test model:
 
    - parent: !uuid f28ec0f8-f3b3-43a0-8af7-79f194b29a2d
 
-Creating objects
-----------------
+Extending objects
+-----------------
+
+The following subsections show how to create completely new, reference and move
+objects using examples of declarative YAML files on
```
Wuestengecko, line 91: I'd rewrite this sentence as "[...] create completely new objects, or reference and move already existing ones [...]". Reads a lot smoother to me.

docs/source/start/declarative.rst
```diff
@@ -153,6 +162,62 @@ functions, which communicate through a functional exchange:
...
+
+.. code-block:: yaml
+
+   - parent: !uuid f28ec0f8-f3b3-43a0-8af7-79f194b29a2d
+     extend:
+       functions:
+         - !uuid 8833d2dc-b862-4a50-b26c-6f7e0f17faef
+
+since the ``functions`` attribute has a parent/children relationship (i.e. the
+:py:class:`~capellambse.model.common.accessors.DirectProxyAccessor` is used).
```
Wuestengecko, line 219: This part of the sentence should be above the code block with the beginning of it, otherwise it looks like it was meant to be another sentence.

Wuestengecko (changes requested):

Wuestengecko (approved):

Issue #193, opened 2022-11-08
Getting properties of associations with py-capellambse


ewuerger:
### Discussed in https://github.com/DSD-DBS/py-capellambse/discussions/190

<div type='discussions-op-text'>

<sup>Originally posted by **khammoutene** November  7, 2022</sup>
Hello,

I'd like to know how I can retrieve the properties of an association between classes like:

- Name
- Kind (e.g. Aggregation)
- Role of each side of the association
- Min./Max. Card. of each side of the association

Thank you</div>

The `Association` is currently not accessable from the `Class` or its `properties` (if modelled via `Association`.

PR #194, opened 2022-11-08
Add `Association`s and all to it

ewuerger:
This PR solves the issue #193 that arose from the discussion #190.

Additionally the [Data Values notebook](https://github.com/DSD-DBS/py-capellambse/blob/add-property-source-from-association/docs/source/examples/03%20Data%20Values.ipynb) was updated with the changes.


vik378 (approved):
nicely done, looks safe enough to try. Maybe we could add also `.all_associations` on the parent data package thing, so that there is also some way to find the structural owner? But this could also be another change for later - if somebody needs all of those now, there is always model.search

Wuestengecko (changes requested):

Wuestengecko:
@ewuerger There are conflicts in the 5.2 test model, but I don't know how to merge that (what to preserve / what to overwrite). Can you please do that? Then I'd say we're probably ready to merge.

Issue #195, opened 2022-11-08
Adding `Class` realizations


ewuerger:
### Discussed in https://github.com/DSD-DBS/py-capellambse/discussions/179

<div type='discussions-op-text'>

<sup>Originally posted by **khammoutene** October 30, 2022</sup>
Hi,

I have a Class defined in a data package at System Analysis level, e.g. ClassSA.

ClassSA is realized by a class in the logical architecture: e.g. ClassLA.

I have "printed" both classes with py-capellambse (using the print statement) but I cannot see in either class reference to the other one: e.g. "ClassLA realizes ClassSA" or "ClassSA is realized by ClassLA".

Am I missing something?

Thank you</div>

Currently there is no way to access the `realized classes` of a `Class`.

PR #196, opened 2022-11-08
Add `realization` attributes to `Class`es

ewuerger:
This PR solves #195 and comes with `realized_classes`, `realizations` (giving access to the new `InformationRealization` object) and `realized_by` attribute for reversed access.


Wuestengecko (changes requested):
Other than the missing tag, it's looking good. :)

Wuestengecko (approved):

PR #200, opened 2022-11-22
Allow ignoring CorruptModelError from duplicate UUIDs

Wuestengecko:
This PR introduces the new (hidden) keyword-only argument `ignore_duplicate_uuids_and_void_all_warranties`. With it set, duplicate UUIDs are logged as warnings instead of raised as exceptions. For additional safety, the `save()` method also requires additional confirmation with the `i_have_a_recent_backup` keyword-only argument.

### DIFF
--- a/capellambse/loader/core.py
+++ b/capellambse/loader/core.py
@@ -148,10 +148,15 @@ class ModelFile:
             return FragmentType.OTHER
 
     def __init__(
-        self, filename: pathlib.PurePosixPath, handler: filehandler.FileHandler
+        self,
+        filename: pathlib.PurePosixPath,
+        handler: filehandler.FileHandler,
+        *,
+        ignore_uuid_dups: bool,
     ) -> None:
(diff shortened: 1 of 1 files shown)


capellambse/loader/core.py
```diff
@@ -408,6 +444,14 @@ def save(self, **kw: t.Any) -> None:
         of ``True`` if you intend to save changes.
         """
         self.check_duplicate_uuids()
+
+        overwrite_corrupt = kw.pop("i_have_a_recent_backup", False)
+        if self.__ignore_uuid_dups and not overwrite_corrupt:
+            raise CorruptModelError(
+                "Refusing to save a corrupt model without having a backup"
+                " (hint: pass i_have_a_recent_backup=True)"
+            )
```
ewuerger, line 453: Nice.

ewuerger (approved):
Really good job.

LGTM.

PR #202, opened 2022-11-23
Fix repr of relations

Wuestengecko:
Relations tried to display the full `repr()` of their source and target, which is not only unnecessarily verbose, but also caused a recursion error. Use the `_short_repr_()` of source and target instead.


ewuerger (approved):
LGTM.

PR #204, opened 2022-11-25
Simplify `_short_repr_` of `ReqIFElement`s and `Relations`

ewuerger:
The title says it all. Pay attention to the added test that is testing the `_short_repr_` appearances in the `__repr__` of certain ReqIFElements instead of testing successful execution via `xtype`.


Wuestengecko:
1. How is the bug you fixed breaking the model? Please explain. As far as I understand, it only affects the visual presentation of a (functioning) model.
2. It seems you accidentally pushed a commit belonging to a different PR onto this one. Please advise.

PR #205, opened 2022-11-25
Fix all pylint complaints

ewuerger:
This should turn the pipeline for #204 green.


Wuestengecko:
> This should turn the pipeline for #204 green.

Then it should either be part of #204 or merged before it, so that #204 can be rebased on top of this and have green CI by itself. Pick one.

Wuestengecko (approved):
Thanks for making the effort! I know this must have been a big push :)

I also took the liberty of hijacking this PR to enable warnings about unused `type:ignore` comments in mypy and clean those up.

PR #206, opened 2022-11-29
feat(declarative-modelling): Add modification from mapping

ewuerger:
In order to handle modify operation from a mapping it is needed to extend `decl._operate_modify` and its `decl._resolve` function.

**WIP**
- The problem left to solve is documented in `accessors.py#L445-452`.
- Rewrite the documentation section


Wuestengecko:
This has been superseded by the `!new_object` support introduced in #387.

PR #207, opened 2022-11-29
Scenarios

Wuestengecko:
This PR introduces support for `Scenario` model objects.


capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+
+
+@c.xtype_handler(None)
+class ExecutionEnd(InteractionFragment):
+    event = c.AttrProxyAccessor[Event](None, "event")
+
+
+@c.xtype_handler(None)
+class FragmentEnd(InteractionFragment):
+    pass
```
ewuerger, line 124: Same inconsistency.

capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+
+class EventOperation(Event):
+    """Abstract super class for events about operations."""
+
+    operation = c.AttrProxyAccessor(c.GenericElement, "operation")
+
+
+@c.xtype_handler(None)
+class ExecutionEvent(Event):
+    pass
```
ewuerger, line 79: All other classes have docstrings instead of the pass statement. Even though the docstring doesn't add much more information. This is inconsistent with the other classes.

capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+
+
+@c.xtype_handler(None)
+class EventSentOperation(EventOperation):
+    pass
```
ewuerger, line 84: Same inconsistency.

capellambse/model/crosslayer/interaction.py
ewuerger:
```suggestion
    """An execution event."""
```

capellambse/model/crosslayer/interaction.py
ewuerger:
```suggestion
    """An event-sent operation."""
```

capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+
+
+@c.xtype_handler(None)
+class EventReceiptOperation(EventOperation):
+    pass
```
ewuerger, line 89: Same inconsistency.

ewuerger:
```suggestion
    """An event-receipt operation."""
```

capellambse/model/crosslayer/interaction.py
ewuerger:
```suggestion
    """An end for a fragment."""
```

capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+
+
+@c.xtype_handler(None)
+class FragmentEnd(InteractionFragment):
+    pass
+
+
+@c.xtype_handler(None)
+class InteractionOperand(InteractionFragment):
+    guard = c.AttrProxyAccessor(capellacore.Constraint, "guard")
```
ewuerger, line 129: No _useless_ docstring.

ewuerger:
```suggestion
    """An interaction-operand."""
    guard = c.AttrProxyAccessor(capellacore.Constraint, "guard")
```

capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+
+
+@c.xtype_handler(None)
+class InteractionState(InteractionFragment):
+    state = c.AttrProxyAccessor(c.GenericElement, "relatedAbstractState")
```
ewuerger, line 134: No _useless_ docstring.

ewuerger:
```suggestion
    """An interaction-state."""
    state = c.AttrProxyAccessor(c.GenericElement, "relatedAbstractState")
```

capellambse/model/crosslayer/interaction.py
```diff
@@ -17,10 +20,125 @@
...
+    function = c.AttrProxyAccessor(c.GenericElement, "relatedAbstractFunction")
+
+
+@c.xtype_handler(None)
+class MessageEnd(InteractionFragment):
+    event = c.AttrProxyAccessor[Event](None, "event")
```
ewuerger, line 140: No _useless_ docstring.

ewuerger:
```suggestion
    """A message-end."""
    event = c.AttrProxyAccessor[Event](None, "event")
```

capellambse/model/modeltypes.py
```diff
@@ -200,3 +200,11 @@ class AggregationKind(_StringyEnumMixin, _enum.Enum):
     ASSOCIATION = _enum.auto()
     AGGREGATION = _enum.auto()
     COMPOSITION = _enum.auto()
+
+
+class ScenarioKind(_StringyEnumMixin, _enum.Enum):
+    UNSET = _enum.auto()
```
ewuerger, line 206: Here is also no _useless_ docstring but other enums except `AggregationKind` also have none.

capellambse/model/common/element.py
```diff
@@ -277,7 +277,11 @@ def _short_repr_(self) -> str:
             mytype = f"Model element ({self.xtype})"
         else:
             mytype = type(self).__name__
-        return f"<{mytype} {self.name!r} ({self.uuid})>"
+        if self.name:
+            name = f" {self.name!r}"
+        else:
+            name = ""
+        return f"<{mytype}{name} ({self.uuid})>"
```
ewuerger, line 284: You don't like the one liner here?

```suggestion
        name = f" {self.name!r}" if self.name else ""
```

ewuerger (changes requested):
Idk why we have so many docstrings in the layer modules since they add literally no value. For consistency (and your convenience) I added suggestions for adding (or exchanging the `pass` statement) with docstrings that follow the pattern. But I think ideally we want to get rid of them at all. What do you think?

Apart from that: LGTM.

capellambse/model/common/element.py
Wuestengecko: No, I don't. I think it unnecessarily obscures the actual program flow.

Wuestengecko:
> Idk why we have so many docstrings in the layer modules since they add literally no value. For consistency (and your convenience) I added suggestions for adding (or exchanging the `pass` statement) with docstrings that follow the pattern. But I think ideally we want to get rid of them at all. What do you think?

I agree they're mostly useless, but having them makes pydocstyle shut up. So let's keep them anyways :)

I think the goal should be to eventually have docstrings everywhere (as in: for everything that's public), and eventually we'll want to make pydocstyle complain loudly about missing docstrings. Currently that's ignored in `.pre-commit-config.yaml`.

PR #208, opened 2022-12-01
Documentation for the low-level API

### LINKED ISSUE #201: Example for low-level API
Hi all
in the README.rst under Current Limitations, there is a link to an example how to use the low level API. Unfortunately it's currently not working. It would be great if you could upload the example, it would make it easier to get into the lower level of the library and maybe also motivate some to contribute ;-)

Wuestengecko:
This PR adds documentation for the low-level API. In doing so, it also improves on related preexisting docstrings in the `capellambse.loader` package, and adds a few minor convenience improvements to the API.

Resolves #201.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -195,7 +195,7 @@ the `low-level API`__.
 
 __ #TODO
 __ #TODO
-__ #TODO
+__ https://dsd-dbs.github.io/py-capellambse/development/low-level-api.html
 
 Also, as we started in mid 2019 and there was no such thing as
 `Python4Capella`__ yet, we are not API compatible with that project. However,
--- a/capellambse/loader/core.py
+++ b/capellambse/loader/core.py
@@ -474,6 +474,11 @@ class MelodyLoader:
 
         This method must be called after adding ``subtree`` to the XML
         tree.
+
+        Parameters
+        ----------
+        subtree
+            The new element that was just inserted.
         """
         try:
             _, tree = self._find_fragment(subtree)
@@ -489,6 +494,11 @@ class MelodyLoader:
 
         This method must be called before actually removing ``subtree``
         from the XML tree.
+
+        Parameters
+        ----------
+        subtree
+            The element that is about to be removed.
         """
         try:
             _, tree = self._find_fragment(subtree)
@@ -521,6 +531,11 @@ class MelodyLoader:
             constraints. If it does not satisfy all constraints (e.g. it
             would be non-unique), a random UUID will be generated as
             normal.
+
+        Returns
+        -------
+        str
+            The new UUID.
         """
 
         def idstream() -> t.Iterator[str]:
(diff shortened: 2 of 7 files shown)


README.rst
```diff
@@ -195,7 +195,7 @@ the `low-level API`__.
 
 __ #TODO
 __ #TODO
```
ewuerger, line 197: What are these for?

capellambse/loader/core.py
```diff
@@ -692,6 +726,13 @@ def find_by_xsi_type(
             A list of XML elements to use as roots for the query.
             Defaults to all tree roots.
         """
+        import warnings
+
+        warnings.warn(
+            "find_by_xsi_type is deprecated, use `iterall_xt` instead",
+            DeprecationWarning,
+        )
```
ewuerger, line 734: Do we need all the hassle with deprecating features of the low-level API? I mean with this PR you kind of make it public, such that we can expect usage of it too. But till then, I'd say we can change things as we go.

docs/source/development/low-level-api.rst
```diff
@@ -0,0 +1,307 @@
...
+   'defend the surrounding area against Intruders'
+   >>> myfunc._element.attrib["name"] = "My Function"
+   >>> myfunc.name
+   'My Function'
+
+Be aware that the XML usually does not explicitly store attributes that are set
+to their default value (as defined by the meta model). In addition to that, the
+high-level API often offers convenience shortcuts and reverse lookups that are
+not directly reflected by XML attributes. This means it can be difficult to
+infer the correct attributes for an object when only looking at the high-level
```
ewuerger, line 230: I think here is a semantical typo. Didn't you mean the low-level API? You praised the convenience shortcuts and reverse lookups for easier access of attributes and their interesting relationships. It doesn't make sense that inference of correct attributes can be difficult.

```suggestion
infer the correct attributes for an object when only looking at the low-level
```

ewuerger (changes requested):
Amazing work. I'm surprised how easy it is for you to find the right words for our work. I think all we need now is to also document the `Accessors` in regards to the meta model implementation.

I've also added lxml to the intersphinx mapping, such that not only explicit external references (there was one to the famous `etree._Element` work but also the type annotations from the sphinx  autodoc extension links to the LXML read the docs.

docs/source/development/low-level-api.rst
Wuestengecko: I mean that, precisely because the high-level API has these additional shortcuts, lookups, etc., it can be difficult to know what to input at the lower level. Could this be phrased a bit better maybe?

capellambse/loader/core.py
Wuestengecko: I primarily threw in the warning there because I was concentrating on the docstrings at the time and didn't want to get distracted. :) Otherwise I probably would have changed it to defer to `iterall_xt` as well.

> with this PR you kind of make it public

It was already part of the public API before (it even shows up in the auto-generated APIDOC), it was just lacking an explanation of the broader concept on how things work together. So I would stick to standard procedures here too.

README.rst
Wuestengecko: What do you mean? These are normal link definitions, just not (yet) pointing anywhere useful. Exactly like the "low-level API documentation" one that I filled in.

ewuerger: Yes and what are we expecting here in the future? In anyway... has nothing to do with this PR.

docs/source/development/low-level-api.rst
ewuerger: Do you mean something like:

> [...] often offers convenience shortcuts, filters and reverse lookups that are
> not directly reflected by XML attributes and may be designed for special use cases.
> Without looking into the rules or declaration in the high-level API it can be difficult to
> infer the correct attributes for an object at the low-level API.

README.rst
Wuestengecko: Just look at the link text in the paragraph right above: For the first one it's `` `high-level API`__ ``, for the second one it's `` `here`__ ``. :)

docs/source/development/low-level-api.rst
Wuestengecko: Yes, that sounds better indeed. I've pushed a new version, can you take a look at it again?

PR #209, opened 2022-12-05
Fix `ElementList.__html__` with `None` members

ewuerger:
When deleting targets of `RequirementOutRelation` or `RequirementIncRelation`s Capella doesn't delete these relations. The leftover and invalid relations caused an exception of requirement display in a jupyter notebook environment. This PR makes this more safe and adds a test for 2 broken relations.

### DIFF
--- a/capellambse/model/common/element.py
+++ b/capellambse/model/common/element.py
@@ -686,7 +686,8 @@ class ElementList(cabc.MutableSequence, t.Generic[T]):
 
         fragments = ['<ol start="0" style="text-align: left;">']
         for i in self:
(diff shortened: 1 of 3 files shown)


Wuestengecko (changes requested):

Wuestengecko:
This needs to be solved differently. `ElementList` is not prepared to handle `None` (as evidenced by the type annotations), and having a `None` anywhere will cause issues in several places. Rather, we need to fix the `RequirementsRelationAccessor` not discard links that are broken anyways.

PR #211, opened 2022-12-07
Improve writing to Capability attributes

Wuestengecko:


capellambse/model/layers/ctx.py
```diff
@@ -113,19 +113,19 @@ class Capability(c.GenericElement):
         fa.FunctionalChain, aslist=c.ElementList
     )
     involved_functions = c.LinkAccessor[SystemFunction](
-        None,  # FIXME fill in tag
-        interaction.XT_CAP2ACT,
+        "ownedAbstractFunctionAbstractCapabilityInvolvements",
+        interaction.AbstractFunctionAbstractCapabilityInvolvement,
```
ewuerger, line 117: It feels a bit weird that we need the tag for the `LinkAccessor`, since this could be inferred from the `_xmltag` class-attribute of the next parameter (`xtype`). But we also want to allow strings there, not only classes and we don't really enforce `_xmltag` to be declared too.

ewuerger (approved):
LGTM.

PR #215, opened 2022-12-15
Internal module restructuring

### LINKED ISSUE #103: Restructuring of internal modules
Some submodules of `capellambse` have evolved a lot since their inception, and now do more tasks than what their name suggests. Additionally, some nested submodules have grown useful enough on their own to warrant them living further up the hierarchy. And finally, some modules redundantly have their parent modules' names in their own name. I therefore suggest the following renames within the `capellambse` package:

- `aird.diagram` 🠪 `diagram._diagram`, exported on `diagram`
- `aird.json_enc` 🠪 `diagram._json_enc`, exported on `diagram`
- `aird.vector2d` 🠪 `diagram._vector2d`, exported on `diagram`
- `aird.capstyle` 🠪 `diagram.capstyle`, maybe with a different name like `default_styles` or `styles`
- `aird.parser` 🠪 `aird`
- `loader.filehandler` 🠪 `filehandler`
- `loader.filehandler.gitfilehandler` 🠪 `filehandler.git`
- `loader.filehandler.git_askpass` 🠪 `filehandler.git.askpass`
- `loader.filehandler.localfilehandler` 🠪 `filehandler.local`

To preserve backwards compatibility, we should put redirections in place for some time, which trigger a `DeprecationWarning` when used. Once sufficient time has passed to allow clients to migrate, these redirections can then be removed.

Summoning @ewuerger and @MoritzWeber0 for their significant contributions to the affected modules.

This issue, if done right, should also allow us to re-enable the `cyclic-import` check in Pylint and not get overwhelmed with output from it.

Wuestengecko:
Resolves #103.

Unfortunately the optimistic statement about Pylint's `cyclic-import` warning does not hold (at least not with this alone), so it remains disabled.

I also placed stubs at the appropriate places, to preserve compatibility with (well-behaved) plugins and give them time to migrate as well. These should be removed after some time again (but hopefully less time than the cache dir migration ;) )


capellambse/aird/__init__.py
```diff
@@ -1,16 +1,246 @@
...
+]
+
+import collections.abc as cabc
+import pathlib
+import typing as t
+import urllib.parse
+
+from lxml import etree
+
+import capellambse
```
ewuerger, line 22: Remove unused import.
```suggestion
```

ewuerger (changes requested):
Please have a look at my additional changes. Apart from that and the minor 2 things:
LGTM.

_**As soon as this is merged, please have a look at [PR#27](https://github.com/DSD-DBS/capellambse-context-diagrams/pull/27) of capellambse-context-diagrams.**_ @Wuestengecko @ewuerger

PR #217, opened 2022-12-16
Selective parsing of AIRD files

Wuestengecko:
This PR aims to reduce the memory usage of various `capellambse` objects during different stages of their life cycle. As a positive side effect, this also greatly reduces the load times for models in the presence of visual representations.

Here are some preliminary measurements made with our production model:

|| `master` | `aird-memory` |
|---|---:|---:|
| Time to prompt | ~18s | 1.4s |
| RSS in prompt | 1480MB | 167MB |
| RSS peak | 1537MB | 400MB |

The biggest hurdle is going to be the AIRD parser, which is currently entirely dysfunctional due to not being able to find its elements. It needs to be adjusted to work with selectively parsed AIRD files.

### DIFF
--- a/capellambse/loader/core.py
+++ b/capellambse/loader/core.py
@@ -132,7 +132,12 @@ class ResourceLocationManager(dict):
 
 
 class ModelFile:
-    """Represents a single file in the model (i.e. a fragment)."""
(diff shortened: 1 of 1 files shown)


Wuestengecko:
Will be implemented as part of model v2.

PR #218, opened 2022-12-27
Enable auditing changes made to a model

ewuerger:
This PR introduces new audit events which allow to detect changes made to the model.

### DIFF
--- a/capellambse/model/__init__.py
+++ b/capellambse/model/__init__.py
@@ -25,7 +25,6 @@ import capellambse
 import capellambse.helpers
 import capellambse.pvmt
 from capellambse import _diagram_cache, filehandler, loader
-from capellambse.loader import xmltools
 
 from . import common, diagram  # isort:skip
 
@@ -64,15 +63,12 @@ class MelodyModel:
         None, cacheattr="_MelodyModel__diagram_cache"
     )
 
-    uuid = xmltools.AttributeProperty(
-        "_element",
+    uuid = common.AttributeProperty(
         "id",
         writable=False,
         __doc__="The unique ID of the model's root element.",
     )
-    name = xmltools.AttributeProperty(
-        "_element", "name", __doc__="The name of this model."
-    )
+    name = common.AttributeProperty("name", __doc__="The name of this model.")
 
     _diagram_cache: filehandler.FileHandler
     _diagram_cache_subdir: pathlib.PurePosixPath
--- a/capellambse/model/__init__.py
+++ b/capellambse/model/__init__.py
@@ -76,6 +76,7 @@ class MelodyModel:
 
     _diagram_cache: filehandler.FileHandler
     _diagram_cache_subdir: pathlib.PurePosixPath
+    _constructed: bool
 
     def __init__(
         self,
@@ -218,6 +219,7 @@ class MelodyModel:
         """
         capellambse.load_model_extensions()
 
+        self._constructed = False
         self._loader = loader.MelodyLoader(path, **kwargs)
         self.info = self._loader.get_model_info()
         self.jupyter_untrusted = jupyter_untrusted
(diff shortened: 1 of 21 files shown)


capellambse/model/__init__.py
```diff
@@ -339,3 +339,7 @@ def search(
     def by_uuid(self, uuid: str) -> common.GenericElement:
         """Search the entire model for an element with the given UUID."""
         return common.GenericElement.from_model(self, self._loader[uuid])
+
+    def __contains__(self, obj: common.GenericElement) -> bool:
+        assert isinstance(obj, common.GenericElement)
+        return obj._model is self
```
ewuerger, line 345: Just an idea: Since I don't want to call `obj._model is model` in the tests. I recognized we don't yet have a quick check for this.

capellambse/model/common/accessors.py
```diff
@@ -23,6 +23,7 @@
 
 import abc
 import collections.abc as cabc
+import copy
```
ewuerger, line 26: I'm not sure if copying the deleted (lost) objects is really needed. If so can we get away without deep-copying?

capellambse/model/common/element.py
```diff
@@ -216,7 +218,7 @@ def __init__(
                         f"Cannot set {key!r} on {type(self).__name__}"
                     )
                 else:
-                    setattr(self, key, val)
+                    self.__setattr__(key, val, audit=audit)
```
ewuerger, line 221: The only change that I don't like so much. But I have no idea how to do it such that we can keep `setattr()`.

capellambse/model/common/element.py
```diff
@@ -225,8 +227,14 @@ def __init__(
     def __getattr__(self, attr: str) -> t.Any:
         raise AttributeError(f"{attr} isn't defined on {type(self).__name__}")
 
-    def __setattr__(self, attr: str, value: t.Any) -> None:
+    def __setattr__(self, attr: str, value: t.Any, audit: bool = True) -> None:
```
ewuerger, line 230: Meh. :(

ewuerger:
There is one WIP commit which keeps me buffled. I want to understand why this doesn't work.

capellambse/model/__init__.py
Wuestengecko: Why would we need to check that `obj._model is model` though? What's the relevant bit in the tests that you're talking about?

capellambse/extensions/reqif/elements.py
```diff
@@ -298,10 +298,17 @@ def __repr__(self) -> str:
...
+        else:
+            value = self.value
+
+        definition_name = re.sub(r"(?<!^)([A-Z])", r" \1", mytype)
         if self.definition is not None:
-            return f"<{mytype} {self.definition.long_name!r} ({self.uuid})>"
-        default_name = re.sub(r"(?<!^)([A-Z])", r" \1", mytype)
-        return f"<{mytype} [{default_name}] ({self.uuid})>"
+            definition_name = self.definition.long_name
+        return f"<{mytype} [{definition_name}] {value!r} ({self.uuid})>"
```
Wuestengecko, line 311: Taking the `repr()` of a formatted date string here makes the output look like it was a `str`. It's not too big of an issue, because it says that it's a `Date` right before, but it could still be confusing. However, we could easily resolve it by taking the `repr()` at the appropriate places during the type checks above and dropping `!r` down here.

capellambse/extensions/reqif/elements.py
```diff
@@ -298,10 +298,17 @@ def __repr__(self) -> str:
 
     def _short_repr_(self) -> str:
         mytype = type(self).__name__
+        if mytype.startswith("Enum"):
+            value = ", ".join(self.values.by_long_name)
+        elif mytype.startswith("Date"):
+            value = helpers.strf_datevalue(self.value)
```
Wuestengecko, line 304: IMO we shouldn't alter the value in any way, which this helper does (it converts it to UTC). We should just use `value.isoformat()` directly.

Wuestengecko (commented):
For the record: I have "taken over" this PR and will be working on it. I'm just taking notes here. :)

PR #222, opened 2023-01-10
Solved bug in ComponentExchange.exchange_items

juanitocalero:
A local list variable in the object was being modified on each call. This caused a bug where code calling this property will produce more elements each time it was called.

### DIFF
--- a/capellambse/model/crosslayer/fa.py
+++ b/capellambse/model/crosslayer/fa.py
@@ -224,11 +224,12 @@ class ComponentExchange(AbstractExchange):
     def exchange_items(
         self,
     ) -> c.ElementList[information.ExchangeItem]:
-        items = self.allocated_exchange_items
-        func_exchanges = self.allocated_functional_exchanges
-        assert isinstance(func_exchanges, cabc.Iterable)
-        for exchange in func_exchanges:
-            items += exchange.exchange_items
+
+        items = c.ElementList(self._model, [], information.ExchangeItem)
+        items.extend(self.allocated_exchange_items)
+        items.extend([item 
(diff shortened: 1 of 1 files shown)


capellambse/model/crosslayer/fa.py
```diff
@@ -224,11 +224,12 @@ def owner(self) -> cs.PhysicalLink | cs.PhysicalPath | None:
...
-        func_exchanges = self.allocated_functional_exchanges
-        assert isinstance(func_exchanges, cabc.Iterable)
-        for exchange in func_exchanges:
-            items += exchange.exchange_items
+
+        items = c.ElementList(self._model, [], information.ExchangeItem)
+        items.extend(self.allocated_exchange_items)
+        items.extend([item 
+            for fexch in self.allocated_functional_exchanges 
+            for item in fexch.exchange_items])
```
Wuestengecko, line 232: Could you please look into setting up the [pre-commit hooks](https://pre-commit.com/) in our working copy? There's some formatting issues here that black would have taken care of. (I recognize we're missing a mention of that in this repo's CONTRIBUTING.)

For now, I've done this; and I've also changed the second part of this function back so it conforms to the Google style (only one `for` statement per comprehension).

Wuestengecko:
Github is being weird today. It's telling me that I can "Add more commits by pushing to the [component-exchange-bug](https://github.com/juanitocalero/py-capellambse/tree/component-exchange-bug) branch on [juanitocalero/py-capellambse](https://github.com/juanitocalero/py-capellambse)", but whenever I try I only get an "access denied" error. I'll just squash this straight in, I guess.

Issue #221, opened 2023-01-09
State Diagram Export Problem


juanvaccarezza:
Hi!
While exporting State Machine Diagrams several problems are detected on transitions:

Image inside capella:
![MSM  test](https://user-images.githubusercontent.com/1491039/211408156-2cdd17fa-776f-4e9c-a5db-cdb8c159604b.png)
Image on jupyter:
![Screenshot 2023-01-09 180609](https://user-images.githubusercontent.com/1491039/211408419-7694d834-cba1-4cb3-b437-62a8b4936a75.png)

Problem 2:
Image inside capella:
![MSM  test2](https://user-images.githubusercontent.com/1491039/211408635-cd4e92b4-fa80-4b00-adbb-8a927bf67c2a.png)
Image on jupyter:
![Screenshot 2023-01-09 180850](https://user-images.githubusercontent.com/1491039/211408793-dd53db86-f3c2-4986-a82e-7d266cde0f32.png)

Best Regards,
Juan

ewuerger:
Hi juanvaccarezza,

I worked on your issue and got to this state:
Capella:
![image](https://user-images.githubusercontent.com/50786483/212666688-ca15fa0d-6fdd-4344-a17b-d6d51944bf18.png)

generated SVG:
![image](https://user-images.githubusercontent.com/50786483/212669285-e781c358-2600-4a5a-9f59-043f46c5d4ab.png)

The rounded corners aren't easy to fix since 2 individual boxes are rendered. Further more pixel perfect routing (start + end points of transitions) are hard to calculate. We don't get all data from the `.aird` files. Please let me know if this is enough for you to work with.

Greetings

ewuerger:
Hi Juan,
sorry for replying so late. The label is taken directly form the `name` attribute of the diagram element in the .aird file. Capella takes the text from the trigger expression. On a `StateTransition` you don't just have a description. Checkout what attributes are available for it [here](https://dsd-dbs.github.io/py-capellambse/code/capellambse.model.crosslayer.html#capellambse.model.crosslayer.capellacommon.StateTransition).

Cheers

juanvaccarezza:
Excellent Thanks!!

Just for the record I was referring to the Trigger Description text box
shown in the dialog when you double click on the transition.

Thanks again,
Juan

ewuerger:
[...]

PR #227, opened 2023-01-16
Fix state diagrams export

ewuerger:
This PR solves #221 by aligning region text correctly and gathering the whole `StateTransition` label text from the `name` attribute on the diagram element in the factory in the `aird.parser module`. The changes to the svg submodule allow for adding class names when we encounter any other elements that need left-alignment of text instead of the default centric placement.


capellambse/svg/decorations.py
```diff
@@ -17,18 +17,22 @@
...
 all_directed_ports = directed_component_ports | function_ports
-start_aligned = {
+start_anchored = {
     "LogicalComponent",
     "LogicalActor",
     "LogicalHumanActor",
     "LogicalHumanComponent",
+    "ModeRegion",
 }
+left_aligned = {"ModeRegion"}
```
Wuestengecko, line 35: As far as I can tell, `StateRegion` has largely the same behavior and formatting as `ModeRegion` (except for background color).

Refer to the two diagrams in test-5.0:

- [MSM] States of Functional Human Being
- [MSM] Weather Modes

Wuestengecko, line 35: Can we change the text alignment setting to be a property in `capstyle.STYLES`?

capellambse/aird/_edge_factories.py
```diff
@@ -470,17 +470,7 @@ def state_transition_factory(seb: C.SemanticElementBuilder) -> diagram.Edge:
...
-        )
-        label = ", ".join(
-            i.get("name", "(unnamed trigger)")
-            for i in triggers
-            if i is not None
-        )
-
-        if guard := _guard_condition(seb, "guard"):
-            label = f"{label} [{guard}]"
+        label = seb.diag_element.get("name", "")
```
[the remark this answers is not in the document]
ewuerger: I didn't know that it is unreliable. So the diagram data in the `.aird` is updated iff the user opened the diagram and saved it again? I'll revert the changes and extend the code from before.

capellambse/svg/decorations.py
ewuerger: Yep. I need to add `StateRegion` too.

ewuerger: We moved all text based decoration rules into `diagram.capstyle.STYLES`. Please have a look again. And tell me if you like this or not.

PR #234, opened 2023-01-24
Model description (complexity) badge

vik378:
This adds a basic summary badge property to a model. 

in response to `model.description_badge` we get an SVG string that provides a visual summary of model composition and key object counts across the engineering layers. 

I wonder if we could also make this a default `display` behavior for the model object itself, since this function call doesnt seem to cost much in terms of resources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,9 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# Mac garbage
+.DS_Store
+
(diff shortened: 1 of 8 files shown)


Wuestengecko (approved):
Aside from the old-style type annotations, here are a few minor issues that I've also addressed. But other than that, it looks fine.

(Obviously ignoring the fact that eventually this should be migrated to use the `capellambse.diagram` facilities for now.)

PR #236, opened 2023-01-26
Fix minor bugs from change audit and requirements merge

ewuerger:
This PR fixes some minor bugs that must have happened during the merge of the requirements into the change-auditing branch.
Somehow the passed element in an `capellambse.delete` audit event wasn't the parent object (as it is documented) but the deleted object itself. Additionally `_operate_modify` would fire an `AttributeError` b/c an `__delete__` was missing on the `AttrProxyAccessor` which is needed for setting a new values-list on an `EnumerationValueAttribute`.


Wuestengecko (changes requested):
We also need new test cases for the bugs you fixed. We're doing test-driven after all, aren't we? :^)

PR #237, opened 2023-01-30
Automatically create a Part for new Components

ewuerger:
This PR implements the `cs.Part` creation of at least one part when creating a new Component. The name of the Part is now correctly inferred from its type.

### DIFF
--- a/src/capellambse/metamodel/cs.py
+++ b/src/capellambse/metamodel/cs.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 import typing as t
 import warnings
 
+from lxml import etree
+
 import capellambse.model as m
 
 from . import capellacore, fa, information, modellingcore
@@ -191,7 +193,8 @@ class Component(
     physical_ports = m.Filter["PhysicalPort"](
         "owned_features", (NS, "PhysicalPort")
     )
-    parts = m.Backref["Part"]((NS, "Part"), "type")
+    owned_parts = m.Filter["Part"]("owned_features", (NS, "Part"))
+    representing_parts = m.Backref["Part"]((NS, "Part"), "type")
     physical_paths = m.Containment["PhysicalPath"](
         "ownedPhysicalPath", (NS, "PhysicalPath")
     )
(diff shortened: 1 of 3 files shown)


Wuestengecko:
Can you rebase on top of latest master (c2e00d4e075701327926795e067b6c6fe0ed9ad5)? That should fix the CI.

capellambse/model/common/element.py
```diff
@@ -199,26 +199,32 @@ def __init__(
...
         if self._xmltag is None:
-            raise TypeError(
-                f"Cannot instantiate {type(self).__name__} directly"
-            )
+            try:
+                self._xmltag = kw.pop("_xmltag")
+            except KeyError as error:
+                raise TypeError(
+                    f"Cannot instantiate {type(self).__name__} directly"
+                ) from error
```
Wuestengecko, line 207:
```suggestion
            except KeyError:
                raise TypeError(
                    f"Cannot instantiate {type(self).__name__} directly"
                ) from None
```

IMO keeping the KeyError just unnecessarily complicates the stack trace without providing additional value. Quite the opposite; it could encourage users to manually pass the `_xmltag` argument (which is definitely not what we want).

tests/test_xlayer_cs.py
```diff
@@ -74,3 +78,45 @@ def test_PhysicalLink_setting_source_and_target(model: MelodyModel):
...
+def test_component_modification_also_modifies_parts(model: MelodyModel):
+    name = "Test"
+
+    comp = model.by_uuid(HOGWARTS_UUID)
+    assert comp.representing_parts
+    comp.name = name
+    comp.allocated_functions.append(model.la.root_function)
+
+    assert (part := comp.representing_parts[0]).name == name
+    assert part.type.allocated_functions == comp.allocated_functions
```
Wuestengecko, line 122: This assertion is wrong, what you really should be checking here is that `part.type == comp`. Or just remove it altogether, because that part of the model isn't touched in the test and therefore doesn't need checking.

```suggestion
```

Wuestengecko (changes requested):

PR #240, opened 2023-02-01
Fix complexity badge display

Wuestengecko:
Some viewers choke on the uppercased `<SVG>` root element. Make it lowercase to fix it.

Also add a heading to the demo notebook, so that it shows up in the index and sidebar.


vik378 (approved):
nice one! who would have guessed that uppercased SVG tag could be a trouble :-)

PR #241, opened 2023-02-02
Basic property value support

Wuestengecko:
Add support for basic property values to the model.

This was originally developed as part of #238.

### DIFF
--- a/capellambse/model/common/element.py
+++ b/capellambse/model/common/element.py
@@ -591,14 +591,18 @@ class ElementList(cabc.MutableSequence, t.Generic[T]):
         return len(self._elements)
 
     @t.overload
-    def __getitem__(self, idx: int | str) -> T:
+    def __getitem__(self, idx: int) -> T:
         ...
 
     @t.overload
     def __getitem__(self, idx: slice) -> ElementList[T]:
         ...
 
-    def __getitem__(self, idx):
+    @t.overload
+    def __getitem__(self, idx: str) -> t.Any:
+        ...
+
+    def __getitem__(self, idx: int | slice | str) -> t.Any:
         if isinstance(idx, slice):
             return self._newlist(self._elements[idx])
         if isinstance(idx, int):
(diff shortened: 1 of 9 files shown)


capellambse/model/crosslayer/capellacore.py
```diff
@@ -26,8 +27,163 @@ class Generalization(c.GenericElement):
...
+    literals = c.DirectProxyAccessor(
+        EnumerationPropertyLiteral, aslist=c.ElementList
+    )
+
+
+class PropertyValue(c.GenericElement):
+    """Abstract base class for PropertyValues."""
+
+    _xmltag = "ownedPropertyValues"
+
```
ewuerger, line 52: ~Do we want a reference back to its group? This is only needed if one is working with `PropertyValue`s via `search` for example.~

```suggestion

group = c.ParentAccessor(c.GenericElement)
```

ewuerger: **Rejected** since this leads to ambiguous returns. Example `StringPropertyValue`'s group can lead to a `EnumerationPropertyLiteral` instead of a `PropertyValueGroup`.

ewuerger (approved):
For my own taste I think we need more work on the notebook to showcase the intent and capabilities of Capella's native property values with capellambse. But this can also happen in a future PR, when one finds the time for it.
Apart from that: Really good job with the nice dict-like behavior for `PropertyValue`s. All of my review comments act like suggestions/thoughts. So nothing that I desperately want to see changed.

LGTM.

capellambse/model/common/element.py
```diff
@@ -707,18 +712,53 @@ def _repr_html_(self) -> str:
...
+        """Get the mapping value from the target object."""
+        assert self.__mapvalue
+        getvalue = operator.attrgetter(self.__mapvalue)
+        return getvalue(obj)
+
+    def _map_setvalue(self, obj: T, value: t.Any) -> None:
+        """Set a new mapping value on the target object."""
+        assert self.__mapvalue
+        key = self.__mapvalue.rsplit(".", maxsplit=1)
+        assert 1 <= len(key) <= 2
```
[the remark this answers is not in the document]
Wuestengecko: Looks like a development leftover. Good catch.

Wuestengecko:
> I think we need more work on the notebook to showcase the intent and capabilities of Capella's native property values

Feel free :)

PR #248, opened 2023-02-21
Add a Gitlab artifact file handler

Wuestengecko:
This file handler uses the Gitlab API to request files from the "artifacts" archive of a recent successful CI job.

### DIFF
--- a/capellambse/filehandler/gitlab_artifacts.py
+++ b/capellambse/filehandler/gitlab_artifacts.py
@@ -1,0 +1,260 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capellambse contributors
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import collections.abc as cabc
+import errno
+import io
+import logging
+import math
+import os
+import pathlib
+import re
+import typing as t
+import urllib.parse
+
+import requests
+
+from capellambse import helpers, loader
+
+from . import FileHandler
+
+LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


capellambse/filehandler/gitlab_artifacts.py
```diff
@@ -0,0 +1,260 @@
...
+                if i >= stop:
+                    return
+
+            match = RE_LINK_NEXT.fullmatch(response.headers["Link"])
+            if not match:
+                break
+            next_url = match.group(0)
+
+    def get_model_info(self) -> loader.ModelInfo:
+        return loader.ModelInfo(branch=self.__branch, url=self.__path)
```
ewuerger, line 241: There is no docstring. But also, what can you write here that isn't already covered by its name.

capellambse/filehandler/gitlab_artifacts.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def open(
+        self,
+        filename: str | pathlib.PurePosixPath,
+        mode: t.Literal["r", "rb", "w", "wb"] = "rb",
+    ) -> t.BinaryIO:
+        path = helpers.normalize_pure_path(filename, base=self.subdir)
```
ewuerger, line 248: No docstring, but also open says it all. Pedantics would ask about a `Raises` section.

ewuerger (approved):
LGTM.

capellambse/filehandler/gitlab_artifacts.py
Wuestengecko: The docstring is inherited from the base class.

PR #261, opened 2023-03-17
Basic support for the Filtering extension

### LINKED ISSUE #252: Introduce basic native support for Filtering extension
Capella has a Filtering extension that provides basic means for model-based variability management. Using this thing you may define on your 150% architecture model what model elements correspond to a "feature" (`FilteringCriterion`), group those features into configurations (`FilteringResult`), have dynamic filtering of views (to preview 100% model on a diagram) and export resulting 100% architecture in a stand-alone model.

You may find more details about the extension here: 
- Introduction: https://github.com/eclipse/capella-filtering/blob/master/plugins/org.polarsys.capella.filtering.doc/html/User%20Manual/Introduction.mediawiki 
- Meta-model: https://github.com/eclipse/capella-filtering/blob/master/plugins/org.polarsys.capella.filtering.doc/html/User%20Manual/Appendix.mediawiki
- Functionality deep-dive: https://github.com/eclipse/capella-filtering/blob/master/plugins/org.polarsys.capella.filtering.doc/html/User%20Manual/Functionalities.mediawiki

Our basic need is to be able to identify if a model element is associated to a FilteringCriterion and list all FilteringCriterions that a model has. The short term goal is to provide enough API support for this so that we could export an HTML page with description of what components / functions / interfaces are available always and what are only available when a specific feature is included (FilteringCriterion is selected).

Wuestengecko:
This PR introduces some basic native support for the Capella Filtering extension. It ain't much, but it's honest work.

Resolves #252.


ewuerger (approved):
LGTM.

PR #265, opened 2023-03-27
Fix exports of diagram images

Wuestengecko:
Two issues were discovered, which could break the automatic generation of diagram images in certain scenarios.

Firstly, the `MelodyLoader.write_tmp_project_dir` function did not correctly reproduce the directory hierarchy within the project. This is especially relevant if the project uses fragment files, as in non-fragmented cases all model files usually are in a single flat folder anyways. This issue caused Capella to not find the associated fragment files, which in turn led to it not seeing all diagrams for export.

Secondly, the generated `.project` file contained the `model.name` as the project name property. This dynamic name caused issues in the diagram image caching function, which expected it to be `main_model`, just like the parent folder's name in the temporary folder. For consistency, the `.project` file now contains the hardcoded project name "main_model".

### DIFF
--- a/capellambse/_diagram_cache.py
+++ b/capellambse/_diagram_cache.py
@@ -313,7 +313,7 @@ class DiagramCache:
                 subprocess.check_call(self._cli_cmd)
             except subprocess.CalledProcessError as err:
                 raise RuntimeError("Failed to update diagram cache") from err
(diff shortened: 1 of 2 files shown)


jamilraichouni (approved):
Great!

PR #266, opened 2023-03-29
Filtering CLI

Wuestengecko:
This PR adds a small CLI that allows to "derive" (= create filtered) projects using native Capella.


capellambse/_native.py
```diff
@@ -0,0 +1,187 @@
...
+        self.workspace = pathlib.Path(ws_path)
+        project = self.model._loader.write_tmp_project_dir()
+        self.project = ctx.enter_context(project)
+        _LOGGER.debug("Project files written to %s", self.project)
+        return self
+
+    def __exit__(self, *args) -> None:
+        assert self.__context is not None
+        self.__context.__exit__(*args)
+        self.__context = self.workspace = self.project = None
```
amolenaar, line 86: Unless there's a strong reason not to, I suggest transforming this to a function with a `@contextlib.contextmanager` decorator. That would provide for more clearly scoped variables (== fewer checks).

capellambse/_diagram_cache.py
```diff
@@ -38,290 +39,185 @@
...
-                )
-        self._cli_common_flags = [
-            "-nosplash",
-            "-consolelog",
-            "-application",
-            "org.polarsys.capella.core.commandline.core",
+                raise ValueError(f"Not found in PATH: {capella}")
+        else:
+            native_args["docker"] = capella
+    assert "exe" in native_args or "docker" in native_args
```
amolenaar, line 88: Could this be moved to a separate function, so it can be tested separately?
It looks like it's mixing different abstraction levels (model export and file path mangling).

amolenaar (approved):
I like that you moved out the "native capella" bits to a separate module. It makes it a lot clearer that the code reaches out to Capella.

The `_diagram_cache` module has some left-overs that deal with the native part. It feels like this code belongs in the `_native` module.

In general the code is good and readable.

I miss changes to tests. Does that mean (1) the code is properly tested and no changes had to be made to the tests, or (2) the code is not tested at all?

capellambse/_native.py
Wuestengecko: The reason here is that we `return self` out of `__enter__`, which for one acts like a callable, but also provides the public attributes `workspace` and `project` – this is needed to copy Capella's output to somewhere safe before destroying the temporary workspace. This is a bit harder to get right if we transform it to a `@contextmanager` function (also taking into account proper type hints). We'd need some class to hold this data and the `__call__` method anyway, so I thought it would make sense to keep everything together and have the class itself act as the context manager.

Wuestengecko: In hindsight, you're right. The benefits from clear scoping outweigh the "keep everything together".

MoritzWeber0:
I've looked into the Gitlab CI template, which wasn't working. `xvfb-run` complained about `Xvfb`, which could not start. The reason for it comes from a combination of the behaviour from `xvfb-run` and the implementation. 

`xvfb-run` and `Xvfb` don't properly clean up the `/tmp` directory, leaving a `.X11-unix` directory behind. I can't exactly say why it happens - the `Xvfb` process seems to terminate properly. 
As we call `xvfb-run` multiple times, `Xvfb` can only start the first time. The second time, it fails, because of an already existing `/tmp/.X11-unix` directory.

As starting up `Xvfb` also takes some time, it makes sense to just start it once for the whole filtering process. As the interface is pretty similar with the diagram cache, I've also added it to our Capella Docker images repository: https://github.com/DSD-DBS/capella-dockerimages/pull/129. Also, it makes setting it up easier for our users as we can hardcode some flags, while the Gitlab CI template in this PR is more flexible.

jamilraichouni:
@Wuestengecko , we need `extension = extension.lower()` in the function `capellambse._diagram_cache._copy_images`.
Otherwise diagrams will not be found, when the user passed a capitalized `image_format` to the method `model.update_diagram_cache()`.

Wuestengecko:
@jamilraichouni 

> diagrams will not be found, when the user passed a capitalized `image_format` to the method `model.update_diagram_cache()`.

Well, _technically_ everything except lowercase violates the API, so _technically_ it's user error at that point :)

```python
    def update_diagram_cache(
        # [...]
        image_format: t.Literal["bmp", "gif", "jpg", "png", "svg"] = "svg",
        # [...]
    ) -> None:
```

But I get your point, I'll add case folding there as well.

capellambse/diagram_cache.py
```diff
@@ -0,0 +1,83 @@
...
+        help="Directory to store the rendered diagrams in",
+        show_defaults=True,
+    )
+    @click.option(
+        "-f",
+        "--format",
+        default="svg",
+        help="Image output format",
+        show_defaults=True,
+    )
```
MoritzWeber0, line 49: I just saw another Github conversation about allowed formats of the diagram cache. I'd use `click.Choice` to enforce the available options with `click`.

jamilraichouni:
> @jamilraichouni
> 
> > diagrams will not be found, when the user passed a capitalized `image_format` to the method `model.update_diagram_cache()`.
> 
> Well, _technically_ everything except lowercase violates the API, so _technically_ it's user error at that point :)
> 
> ```python
>     def update_diagram_cache(
>         # [...]
>         image_format: t.Literal["bmp", "gif", "jpg", "png", "svg"] = "svg",
>         # [...]
>     ) -> None:
> ```
> 
> But I get your point, I'll add case folding there as well.

Uups :-) Well yes, you are right. Anyway it is more safe to lower case it.

capellambse/diagram_cache.py
Wuestengecko: The idea behind allowing any string here was to later let Capella decide whether it knows that format or not; this way Capella can add other image formats in the future without us having to explicitly add them to the Choice options. Do you think that makes sense? I'm also fine with switching to Choice for faster (and probably more useful) error messages.

jamilraichouni: I'd hard code the choices following Moritz' proposal. I don't expect Obeo to do something here. And if they really add/ remove an image format it is a very tiny and quick change we can apply on our side.

MoritzWeber0: @Wuestengecko I can understand the point, but the usability improvement (fail fast principle and better error message, the Capella error message is hard to read) outweighs the chance that there will a new image format.

capellambse/_diagram_cache.py
```diff
@@ -27,301 +28,196 @@
...
+    for diagram in sorted(index, key=operator.itemgetter("name")):
+        uuid = E.span(diagram["uuid"], {"class": "uuid"})
+        if diagram["success"]:
+            href = f"{diagram['uuid']}.{extension}"
+            label = E.a({"href": href}, diagram["name"])
+        else:
+            label = E.span({"class": "missing"}, diagram["name"])
+        ol.append(E.li(label, " ", uuid))
+
+    dest.joinpath("index.json").write_text(json.dumps(index))
```
jamilraichouni, line 222:
```suggestion
    dest.joinpath("index.json").write_text(json.dumps(index, indent=2))
```
Let us write a human-readable file here. A oneliner is useless when one needs to read it. With syntax highlighting enabled it takes quite a while to format the file with a LSP. That can be annoying.

jamilraichouni: When we talk about 22kB or even 20 times that size I do not see the point to discuss network traffic here.
When we want to check a produced JSON file (a scenario I expect to happen) it makes sense, that we can just read it w/o processing.
I totally agree (minifying stuff) when the size requires it. Here, the files will be very small (even with > 1,000 diagrams).
I formatted with the attached LSP `vscode-json-language-server` and the command lua vim.lsp.buf.format().
takes long for an index.json that includes diagram element information having a size of 136 kB and 5,400 lines (when formatted)

PR #30, opened 2021-11-23
Fix port exchange items

vik378:
implemented fix for #28

PR #268, opened 2023-03-30
fix(svg): Draw symbols and circles lastly

ewuerger:
As can be seen in the SVG from [context-diagrams #30](https://github.com/DSD-DBS/capellambse-context-diagrams/pull/30) exchanges might overlap ports. This is fixed by introducing a simple order in drawing objects on the canvas.

The new version of the SVG looks like this:
image

### DIFF
--- a/capellambse/svg/generate.py
+++ b/capellambse/svg/generate.py
@@ -67,7 +67,14 @@ class SVGDiagram:
         objects: cabc.Sequence[ContentsDict],
     ) -> None:
         self.drawing = Drawing(metadata)
+        draw_last = list[ContentsDict]()
         for obj in objects:
+            if obj["type"] in {"symbol", "circle"}:
+                draw_last.append(obj)
+            else:
+                self.draw_object(obj)
+
+        for obj in draw_last:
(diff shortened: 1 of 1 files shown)


Wuestengecko:
Changing the drawing order could negatively impact our in-house renderings of Capella diagrams (yes, I know, they're not exactly awesome to begin with...). I vaguely recall us having a similar issue during early development of capellambse's SVG exporter, and it was fixed by drawing everything in the exact order that it came in from the AIRD. Did you look at anything in that direction?

This could also be fixed by context-diagrams drawing the lines only up to the edge of the port box instead of connecting them together in one point. This is exactly what the `capellambse.aird` parser does. [The `Box` type has a method that could help with that](https://github.com/DSD-DBS/py-capellambse/blob/803d131218bd4b881fb84dcba0c9cc42acb44420/capellambse/diagram/_diagram.py#L242-L267). IMO this would be the cleaner approach, but would it also be viable? What would the implementation effort be on the context-diagrams side?

ewuerger:
This was done in context-diagrams and can therefore be closed.

PR #269, opened 2023-04-04
Add model arguments to Gitlab CI template

MoritzWeber0:
Adds an option in the Gitlab CI template for the model badge to pass additional arguments to the MelodyModel class. Users can set the `CAPELLAMBSE_MODEL_KWARGS` environment variable to a json dict with additional arguments.

### DIFF
--- a/ci-templates/gitlab/README.rst
+++ b/ci-templates/gitlab/README.rst
@@ -20,6 +20,7 @@ Please add the following section to your ``.gitlab-ci.yml``:
 
   variables:
     ENTRYPOINT: test/test.aird # Entry point to the .aird file of the model (relative from root level of the repository)
+    CAPELLAMBSE_MODEL_KWARGS: '{"ignore_duplicate_uuids_and_void_all_warranties": true}' # Additional arguments as json, passed to the MelodyModel class.
 
   # The following section is only needed if you want to change the output filename or commit message
   generate-model-badge:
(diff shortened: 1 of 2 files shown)


Wuestengecko:
IMO we should rather make this a proper CLI which uses `capellambse.cli_helpers.ModelCLI`. (See https://github.com/DSD-DBS/py-capellambse/pull/266/files#diff-fc81a7ffb4e50e4a786e8dd4b8e56059ad55514348c731b3a6bb4485a5ff1c7aR24-R65 for an example.)

This allows to pass JSON as `ENTRYPOINT` like

```yaml
ENTRYPOINT: '{"path": "some.aird", "additional_args": false}'
```

And it would make the CI template be a simple `python -m capellambse.some_module -m "$ENTRYPOINT" -o "$OUTPUTFILE"`

grafik

MoritzWeber0:
> IMO we should rather make this a proper CLI which uses `capellambse.cli_helpers.ModelCLI`. (See https://github.com/DSD-DBS/py-capellambse/pull/266/files#diff-fc81a7ffb4e50e4a786e8dd4b8e56059ad55514348c731b3a6bb4485a5ff1c7aR24-R65 for an example.)
> 
> This allows to pass JSON as `ENTRYPOINT` like
> 
> ```yaml
> ENTRYPOINT: '{"path": "some.aird", "additional_args": false}'
> ```
> 
> And it would make the CI template be a simple `python -m capellambse.some_module -m "$ENTRYPOINT" -o "$OUTPUTFILE"`
> 
> grafik

I agree. For consistency with the CI templates in the [capella-dockerimages](https://github.com/DSD-DBS/capella-dockerimages) repository, I'd integrate the model validation first. 
When the model validation is integrated, we can define py-capellambse specific entrypoints.

MoritzWeber0:
Let's close this one for now.

PR #277, opened 2023-04-24
Show namespace warning instead of error on mismatch

dominik003:
This pull request refactors the `add_namespace` method in the loader module to improve namespace handling. Specifically, it modifies the behavior of the method to log a warning instead of raising an error when an already registered URI is detected. This change enables users to apply a declarative YAML to a Capella 6.0.0 model, which was previously not possible due to the strict error handling in this method. With this change, the URI for a given namespace can be updated if necessary, without causing an error to be raised. Overall, this improves the robustness and usability of the code.

### DIFF
--- a/capellambse/loader/core.py
+++ b/capellambse/loader/core.py
@@ -260,9 +260,10 @@ class ModelFile:
         fragroot = self.root
         olduri = fragroot.nsmap.get(name)
         if olduri is not None and olduri != uri:
-            raise ValueError(
-                f"Namespace {name!r} already registered with URI {olduri!r}"
+            LOGGER.warning(
+                "Namespace %r already registered with URI %r", name, olduri
             )
+            return
         if uri == olduri:
             return
 


capellambse/loader/core.py
```diff
@@ -260,8 +260,8 @@ def add_namespace(self, name: str, uri: str) -> None:
         fragroot = self.root
         olduri = fragroot.nsmap.get(name)
         if olduri is not None and olduri != uri:
-            raise ValueError(
-                f"Namespace {name!r} already registered with URI {olduri!r}"
+            LOGGER.warning(
+                "Namespace %r already registered with URI %r", name, olduri
```
Wuestengecko, line 264: > Specifically, this change allows the URI for a given namespace to be updated if necessary, without causing an error to be raised.

Wouldn't this actually cause even more trouble, by putting e.g. a 5.0 namespace into a 6.0 model? Did you try loading such a model in Capella after this change?

Wuestengecko: So, I just tried it, and yes: If you inject a 5.0 namespace into a 6.0 model, the entire model breaks. So we at least need to `return` early here to prevent overwriting a good namespace with a bad one.

Wuestengecko (changes requested):

capellambse/loader/core.py
dominik003: Yes, I agree that overwriting the existing namespace is not the correct way here. I added a return, so in the case that an namespace exists, it is no longer overwritten.

PR #280, opened 2023-04-27
Lock the repo during calls to git

Wuestengecko:
This fixes an issue where the invoked git process would return an error if a parallel process was already performing some task within the repo (like fetching updates). Instead, we now use an additional `capellambse.lock` file within the repo directory which we can `flock` to make sure that only one git instance will be called at a time.

This `flock` based approach works regardless of which parallelization method is used, i.e. thread-based and process-based both work fine.

### DIFF
--- a/capellambse/helpers.py
+++ b/capellambse/helpers.py
@@ -6,14 +6,19 @@ from __future__ import annotations
 
 import collections
 import collections.abc as cabc
+import contextlib
+import errno
 import functools
 import html
 import importlib.resources as imr
 import itertools
+import logging
 import math
 import operator
 import pathlib
 import re
+import sys
+import time
 import typing as t
 
 import lxml.html
@@ -24,6 +29,13 @@ from PIL import ImageFont
 
 import capellambse
 import capellambse._namespaces as _n
+
+if sys.platform.startswith("win"):
+    import msvcrt
+else:
+    import fcntl
+
+LOGGER = logging.getLogger(__name__)
 
 ATT_XT = f"{{{_n.NAMESPACES['xsi']}}}type"
 FALLBACK_FONT = "OpenSans-Regular.ttf"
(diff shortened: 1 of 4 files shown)


capellambse/helpers.py
```diff
@@ -116,6 +127,50 @@ def normalize_pure_path(
...
+
+else:
+
+    @contextlib.contextmanager
+    def flock(file: pathlib.Path) -> t.Iterator[None]:
+        file = file.resolve()
+        with file.open("wb") as lock:
+            try:
+                fcntl.flock(lock, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except OSError as err:
```
MoritzWeber0, line 164: Can't we just use `BlockingIOError` here? `errno.EAGAIN` should always fall under `BlockingIOError` and all other errors are raised anyway.

MoritzWeber0 (commented):
Only one short comment, I'm not ready with the review yet. Most likely, I will not have time before Monday for a complete review, sorry.

capellambse/helpers.py
Wuestengecko: You're right, good catch!

tests/test_helpers.py
```diff
@@ -93,3 +97,72 @@ def test_flatten_html_formats_unordered_lists(
     input: str, expected: str
 ) -> None:
     assert helpers.flatten_html_string(input) == expected
+
+
+if sys.platform.startswith("win"):
+    import msvcrt
+
```
ewuerger, line 104: To spare the indentation one can use `@pytest.mark.skipif(sys.platform.startswith("linux"), reason="does not run on linux")`. The other test could use `@pytest.mark.skipif(sys.platform.startswith("win"), reason="does not run on windows")`. The import would happen in the body of the test function.

ewuerger (changes requested):
Looks good. LGTM.

PR #286, opened 2023-05-09
Improve workflow triggers

Wuestengecko:
This PR improves our CI workflows a bit:

- Triggers for Code QA were fixed to properly run for incoming pull requests from other repos.
- The linting job (pre-commit) was moved to the build/test/publish workflow file, because it kinda belongs there anyways (we want to lint whenever we test, so it makes sense to have both in one workflow).
- The documentation workflow now only runs on master by default (to update the web docs), but can be manually triggered to build the archive for inspection/debugging.

Most / all of these changes should also eventually propagate to our other public repos, wherever the same workflow definitions are used. Summoning @MoritzWeber0 and @ewuerger, maintainers of such repos, for their valuable opinions.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,7 +4,9 @@
 name: Docs
 
 on:
+  workflow_dispatch:
   push:
+    branches: [master]
 
 jobs:
   sphinx:


.github/workflows/code-qa.yml
```diff
@@ -1,12 +1,14 @@
...
-name: Build
+name: Code quality assurance
 
 on:
+  workflow_dispatch:
   push:
-    branches: ["*"]
+    branches: [master]
     tags: ["v*.*.*"]
+  pull_request:
```
Wuestengecko, line 11: Do we want to keep the `push` event trigger unbounded _in addition to_ `pull_request`, as it is done in [capella-collab-manager](https://github.com/DSD-DBS/capella-collab-manager/blob/main/.github/workflows/lint.yml)? This causes the workflow to run twice for every push to an open PR, however it would make it easier to try out experimental changes in the CI (which is probably most interesting for unit tests on other OSes). As a middle ground between those, I have added the `workflow_dispatch` trigger to the code-qa workflow, which allows to manually run a workflow on demand.

MoritzWeber0 (dismissed):
LGTM. `workflow_dispatch` is great, I'll consider adding it the the Capella Collaboration Manager and Capella Dockerimages repositories too.

PR #302, opened 2023-06-15
Factor out HTML fragment processing

ewuerger:
Helps modifying Markup HTML arbitrarily.

### DIFF
--- a/capellambse/helpers.py
+++ b/capellambse/helpers.py
@@ -367,6 +367,46 @@ def repair_html(markup: str) -> markupsafe.Markup:
     markup
         The repaired markup.
     """
+
+    def cb(node: etree._Element) -> None:
+        for k in list(node.keys()):
+            if ":" in k:
+                del node.attrib[k]
+
+    return process_html_fragments(markup, cb)
+
+
+def process_html_fragments(
+    markup: str, node_callback: cabc.Callable[[etree._Element], None]
+) -> markupsafe.Markup:
+    """Repair and modify HTML markup.
+
+    The original markup, which can be an HTML fragment (without a root
+    element), is parsed and processed, and then reassembled into a
+    Markup instance. If the original markup contained any errors or
+    inconsistencies, these are repaired in the returned Markup instance.
+
+    Parameters
+    ----------
+    markup
+        The markup string to modify.
+    node_callback
+        A callback function to process each node in the parsed markup.
+        The function should accept a single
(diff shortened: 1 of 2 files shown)


capellambse/helpers.py
```diff
@@ -367,6 +367,40 @@ def repair_html(markup: str) -> markupsafe.Markup:
...
+        for k in list(node.keys()):
+            if ":" in k:
+                del node.attrib[k]
+
+    return handle_html(markup, cb)
+
+
+def handle_html(
+    markup: str, node_handler: cabc.Callable[[etree._Element], None]
+) -> markupsafe.Markup:
```
Wuestengecko, line 381: When I worked on this with you, I intended these to be preliminary names so we could concentrate on getting the code working and not waste time thinking about what to name this function that hasn't even been written yet.

`handle_html` says basically nothing, and I'm also not really happy with the `node_handler` argument name. What do you think about this?

```suggestion
def process_html_fragments(
    markup: str, node_callback: cabc.Callable[[etree._Element], None]
) -> markupsafe.Markup:
```

capellambse/helpers.py
```diff
@@ -367,6 +367,40 @@ def repair_html(markup: str) -> markupsafe.Markup:
...
+    """Modify HTML markup using a provided node handler function.
```
Wuestengecko, line 382: The docstring needs to mention these important factoids:

- This accepts and returns an HTML fragment (i.e. there doesn't have to be a surrounding root element)
- More than the first element handed to the callback may have no parent (implied by the previous point, but IMO we should highlight it by mentioning it explicitly)
- The callback won't be called for leading text if there is any, and so has no ability to influence it

capellambse/helpers.py
```diff
@@ -367,6 +367,40 @@ def repair_html(markup: str) -> markupsafe.Markup:
...
+        format.
+    node_handler
+        A function that handles nodes in the parsed markup. The function
+        should accept an ``etree._Element`` instance. This function is
+        applied to each node in the parsed markup.
+
+    Returns
+    -------
+    markupsafe.Markup
+        The modified markup.
```
Wuestengecko, line 402:
```suggestion
        The processed markup.
```

Wuestengecko (changes requested):
Looks generally good, I'm just not entirely happy with the API yet.

Also, you think you could whip up a few quick unit tests?

PR #313, opened 2023-07-19
Display name of element in `__html__` heading

MoritzWeber0:
In the `__html__` representation, which is used for Jupyter, the name of the type of the element was displayed.

While the type is relatively uninteresting for the user, especially because the xtype is displayed right after it, the name of the element itself was missing completely in the representation.

This commit replaces the type name with the element if a element name is available.

State before this commit: 
image

State after this commit: 
image

### DIFF
--- a/capellambse/model/common/element.py
+++ b/capellambse/model/common/element.py
@@ -332,7 +332,7 @@ class GenericElement:
             fragments.append("<h1>Model element")
         else:
             fragments.append("<h1>")
-            fragments.append(escape(type(self).__name__))
+            fragments.append(escape(self.name or type(self).__name__))
         fragments.append(' <span style="font-size: 70%;">(')
         fragments.append(escape(self.xtype))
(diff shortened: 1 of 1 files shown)


Wuestengecko (approved):
Good idea! Although technically, the name _was_ already shown in the `__html__` output, just in small as part of the big list of attributes (where it's easy to overlook it, admittedly). I took the liberty of fixing that in the commit message, and I also slightly shortened the code while I was at it.

LGTM.

MoritzWeber0:
> Good idea! Although technically, the name _was_ already shown in the `__html__` output, just in small as part of the big list of attributes (where it's easy to overlook it, admittedly). I took the liberty of fixing that in the commit message, and I also slightly shortened the code while I was at it.

Indeed. It was so small I must have missed it :D 
And I learned that the Jupyter search doesn't scan the output section.

PR #316, opened 2023-07-21
Fix diagram rich display in JupyterLab

Wuestengecko:
JupyterLab (unlike the legacy "Jupyter notebook server") prefers JSON over SVG for display purposes. Capellambse used to expose its internally used JSON format in the diagrams' `_repr_mimebundle_()` method. This caused JupyterLab to not show diagrams properly when using `display()` or when a Diagram instance was the result of a cell.

Furthermore, the way that the `_repr_mimebundle_()` method returned its `application/json` mimetype (encoded as a string, instead of as regular Python dict) was never supported to begin with.

This PR fixes both of these issues by no longer exposing this internal format publicly. This means that:

- `_repr_mimebundle_()` will no longer contain `application/json`
- `AbstractDiagram.render()` will no longer accept `json` and `json_pretty` as format argument
- The `AbstractDiagram.as_json` and `as_json_pretty` auto-generated properties are no longer present and will raise AttributeErrors

### DIFF
--- a/capellambse/model/diagram.py
+++ b/capellambse/model/diagram.py
@@ -594,31 +594,6 @@ class SVGInHTMLIMGFormat:
         return markupsafe.Markup(f'<img src="{payload}"/>')
 
 
-class JSONFormat:
-    filename_extension = ".json"
-
-    @staticmethod
-    def convert(dg: diagram.Diagram) -> str:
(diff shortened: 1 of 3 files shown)


ewuerger (approved):
LGTM.

PR #317, opened 2023-07-25
More missing CEX attributes

Wuestengecko:
Adds realizing/realized links between CEX, and allows multiple CEX → PhysicalPath allocations.


ewuerger (approved):
LGTM.

PR #321, opened 2023-08-07
feat(model): Resolve diagram-links in descriptions

ewuerger:
## Resolving references to diagrams in descriptions automatically

The `properties.HTMLAttributeProperty` is only ever used for the description of any `GenericElement` or for the text of any `reqif.ReqIFElement`. I.e. when ever a rich-text/HTML attribute is set. In there a reference to a diagram object is now resolved on requesting the attribute, references to non-diagram objects and all other content remains as is.

Example from the test model:
image

Thanks to @Paula-Kli for helping during the implementation.

### DIFF
--- a/capellambse/helpers.py
+++ b/capellambse/helpers.py
@@ -43,6 +43,7 @@ RE_VALID_UUID = re.compile(
     r"[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}",
     re.IGNORECASE,
 )
+RE_VALID_DIAGRAM_UID = re.compile(r"^_[A-Za-z0-9+/]{22}$")
 LINEBREAK_AFTER = frozenset({"br", "p", "ul", "li"})
 TABS_BEFORE = frozenset({"li"})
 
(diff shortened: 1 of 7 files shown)


vik378:
We should not implement this functionality. This will result in bloated descriptions and also deviates from the original intention of a content author. Please do not merge.

ewuerger:
> I actually started to think that capellambse might not be the right place for this functionality after all. Especially now that I'm seeing how much logic this ends up being, I kinda don't really want to do so much on (what the user sees as) a simple attribute access.

I think so too. The feature that I implemented here actually works fine on the get side (imho), but the set side makes some too specific assumptions which restrict the user in the options on images inside of `HTMLAttributeProperty`s. As you mentioned Inline images which are not meant to be diagrams would break. I could work on distinguishing those... but this feels plain wrong in capellambse. This is behavior which should be defined in a derivative product, if at all.

> We should not implement this functionality. This will result in bloated descriptions and also deviates from the original intention of a content author. Please do not merge.

I guess your comment is directed to the `__set__` part, else I don't agree. Evaluating the idea with regards to Change-Control this would induce a nightmare since now dependencies between diagrams and model-objects from descriptions would be created... by capellambse and not Capella. Not cool. Ready to close this PR when you are @vik378 @Wuestengecko

vik378:
I think get part is flawed too, @ewuerger - as an author, I'd like to mention a diagram, like "see more about the thing in action in Scenario1 and Scenario2" - but I don't want to show those in my description. We are creeping into templating - lets not do it, there is a proper app for that

ewuerger:
With all participants to agree we closed this PR.

PR #322, opened 2023-08-11
feat(model): Add `input/output_exchanges` to `fa.Function`

ewuerger:
To be on par with `inputs/outputs` on `oa.OperationalActivity`s these
two properties were added.

### DIFF
--- a/capellambse/model/crosslayer/fa.py
+++ b/capellambse/model/crosslayer/fa.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 import collections.abc as cabc
 import typing as t
 
+from lxml import etree
+
 from .. import common as c
 from .. import modeltypes
 from . import capellacommon, capellacore, information, interaction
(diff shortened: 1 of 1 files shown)


Wuestengecko:
We keep adding more and more special cases like this. I think we should rather look at how we could take a more generalized approach, to make it easy for users to express the exact relationships they're interested in.

I see a case to make some generalized filtering and mapping methods on ElementList, which would transform the list by accessing the given attribute on each item. This example could then be written similar to:

```python
my_func = model.search("LogicalFunction")[0]
exc = my_func.inputs._the_mapper_function_("exchanges")
# exc now contains the same as this PR's `my_func.input_exchanges`
```

Of course the downside is that this cannot be statically type annotated (to the best of my knowledge), so the static type of `exc` would simply be `ElementList[Any]`.

This is just a rough idea of course, but what do you think? Does it make sense to explore this idea further?

Wuestengecko:
Obsolete. capella2polarion has since added support for dotted notation of Capella attributes, which is a more generalized approach than adding the needed attributes to the metamodel.

PR #323, opened 2023-08-15
feat: Add white background if needed to diagram cache svg

Paula-Kli:
This PR adds the option to add white background to the diagram cache via the cli

Example in green to show the extend of the background and that it is only around the diagram
Bildschirmfoto 2023-08-15 um 16 30 00

### DIFF
--- a/capellambse/_diagram_cache.py
+++ b/capellambse/_diagram_cache.py
@@ -62,6 +62,7 @@ def export(
     format: str,
     index: bool,
     force: t.Literal["exe", "docker", None],
+    background: bool,
 ) -> None:
     if not isinstance(
         getattr(model, "_diagram_cache", None), local.LocalFileHandler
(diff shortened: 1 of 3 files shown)


capellambse/_diagram_cache.py
```diff
@@ -62,6 +62,7 @@ def export(
     format: str,
     index: bool,
     force: t.Literal["exe", "docker", None],
+    background: t.Optional[bool],
```
Wuestengecko, line 65: There's no difference between the "False" and the "None" case, so there's no need to distinguish between them.

```diff
-    background: t.Optional[bool],
+    background: bool,
```

capellambse/model/__init__.py
```diff
@@ -573,6 +573,7 @@ def update_diagram_cache(
             format=image_format,
             index=create_index,
             force=force,
+            background=None,
```
Wuestengecko, line 576: This would also make for a nice (optional) method argument, which simply defaults to the current behavior, i.e. "False".

Wuestengecko (commented):
Looks good, only some minor style improvements.

PR #327, opened 2023-08-21
Fix creation of `Relation`s from a `Requirement`

ewuerger:
This is needed to bring our RequirementsBot into working state again. 

While fixing the creation of `Relation`s from the `reqif.Requirement` side, what about the other side? So creating `CapellaOutgoingRelation`s from any model object to a `reqif.Requirement` can't be done via capellambse right now, or? I can't find a `.relation` attribute on `GenericElement`. Also there are 2 ways of creating `Relation`. Both lead to different places of where the `Relation` XML element is written to: Either underneath the model object or underneath the Requirement. The GUI accepts both, so I'd root for capellambse creating `Relation`s just underneath the `reqif.Requirement` to keep the reqif content compact and bundled inside the `RequirementsModule`. The latter hasn't been done for now, and is the main factor why this PR is a DRAFT.

### DIFF
--- a/capellambse/extensions/reqif/_capellareq.py
+++ b/capellambse/extensions/reqif/_capellareq.py
@@ -251,7 +251,7 @@ class RequirementsRelationAccessor(
                 f" {type(target).__name__}"
             )
         else:
-            return rq.InternalRelation
+            return CapellaIncomingRelation
 
 
 class RelationsList(c.ElementList["rq.AbstractRequirementsRelation"]):
--- a/tests/test_reqif.py
+++ b/tests/test_reqif.py
@@ -336,6 +336,38 @@ class TestRequirementRelations:
         )
         assert {i.uuid for i in relations} == target_uuids
 
+    @pytest.mark.parametrize(
+        "obj_uuid,target_uuid,expected_type",
+        (
+            pytest.param(
+                "0a9a68b1-ba9a-4793-b2cf-4448f0b4b8cc",
+                "3c2d312c-37c9-41b5-8c32-67578fa52dc3",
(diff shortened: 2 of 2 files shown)


Wuestengecko:
> While fixing the creation of `Relation`s from the `reqif.Requirement` side, what about the other side? So creating `CapellaOutgoingRelation`s from any model object to a `reqif.Requirement` can't be done via capellambse right now, or?

Correct, currently there's no API for this. Do we need it? Then we need to figure out how that API should look.

> I can't find a `.relation` attribute on `GenericElement`.

`relation` would be a far too generic name anyways, it would at least have to be `requirement_relations` or something like that to indicate what to expect on the other side.

> Also there are 2 ways of creating `Relation`. Both lead to different places of where the `Relation` XML element is written to: Either underneath the model object or underneath the Requirement.

That doesn't sound right. Can you give me some example code?

> The GUI accepts both

The GUI can sometimes be very forgiving, but that doesn't mean what we do is correct or even makes sense. A better question would be: Is it possible to do the same with (only) the GUI?

ewuerger:
To clarify: The difference between these Relations isn't just their position in the XML. They have different xtypes and their source and target differ. `CapellaOutgoingRelation`s underneath a non-`Requirement` element have a `Requirement` referenced in their `target` attribute. The `source` attribute does always reference its parent model element. For now the modification of `CapellaOutgoingRelation`s can't be done with py-capellambse, and there is no intention to enable this before we have a legitimite use case.

Therefore I change this into a non-Draft PR.

Wuestengecko (approved):

Issue #333, opened 2023-09-20
PA demo notebook uses private `_filter` method (and fails)


Wuestengecko:
This (private) method used to be used by the implementation of `by_X` and `exclude_Xs`, but has since been removed. That broke the PA demo notebook.

The notebook wants to filter an ElementList with a lambda, which is currently simply not supported - we should introduce proper support for this use case.

PR #334, opened 2023-09-20
Temporary fix of PA Notebook

dahbar:
Temporary fix of PA Notebook while [this issue](https://github.com/DSD-DBS/py-capellambse/issues/333) is resolved.


Wuestengecko:
This will be directly fixed in a more permanent way by implementing the missing functionality. Thanks anyway :)

PR #341, opened 2023-09-27
Don't use a stylesheet in SVGs

dahbar:
- [x] Remove SVG Stylesheet
- [x] Add test coverage


tests/test_svg.py
```diff
@@ -130,273 +130,19 @@ def test_diagram_saves(self, tmp_svg: SVGDiagram) -> None:
...
-                            raise NotImplementedError
+        style_ = tree.get("style")
+        assert style_ == (
+            """shape-rendering: geometricPrecision; font-family:
+                'Segoe UI'; font-size: 8pt; cursor: pointer;""".replace(
+                "\n", ""
+            )
+            .replace("   ", "")
+            .strip()
+        )
```
Wuestengecko, line 145: I don't think it makes sense to check for the exact string here, because the contents are static and can only change by someone deliberately doing so – in which case that someone has to change it twice for no good reason. A more useful test here would be "is something there", i.e. implicit truthiness, to check that it wasn't accidentally deleted completely:

```suggestion
        style_ = tree.get("style")
        assert style_
```

tests/test_svg.py
```diff
@@ -130,273 +130,19 @@ def test_diagram_saves(self, tmp_svg: SVGDiagram) -> None:
...
+    def test_base_css_styles(self, tmp_json: pathlib.Path) -> None:
         tree = etree.fromstring(
             SVGDiagram.from_json_path(tmp_json).to_string()
         )
-        style_ = tree.xpath(
-            "/x:svg/x:defs/x:style",
-            namespaces={"x": "http://www.w3.org/2000/svg"},
-        )[0]
-
-        stylesheet = cssutils.parseString(style_.text)
```
Wuestengecko, line 374: These tests were the only place where we used `cssutils` - now that we don't have a stylesheet to test anymore, we can get rid of that dependency entirely.

capellambse/svg/drawing.py
```diff
@@ -451,38 +423,49 @@ def draw_object(self, obj: cabc.Mapping[str, t.Any]) -> None:
...
             **{
-                k[len("text_") :]: v
-                for k, v in obj.get("style", {}).items()
-                if k.startswith("text_")
+                k[5:]: v for k, v in my_styles.items() if k.startswith("text_")
             },
         )
 
-        self.obj_cache[mobj["id"]] = obj
+        self.obj_cache[mobj["id"]] = mobj
```
Wuestengecko, line 468: This `self.obj_cache` seems to no longer be used at all. It's only written to, but never read anywhere. While we're cleaning up here anyways, might as well remove that too.

capellambse/svg/style.py
```diff
@@ -261,6 +256,9 @@ def _style_name(self, attr: str) -> str:
 
         return "_".join((self._prefix, attr))
 
+    def __contains__(self, obj: t.Any) -> bool:
+        return hasattr(self, obj)
+
 
 class Style(base.BaseElement):
```
Wuestengecko, line 263: This `Style` class is no longer used either, it can be removed too. :)

Wuestengecko (changes requested):
Looks quite good already, only some minor issues.

After these, we'll also need to deprecate the `jupyter_untrusted` argument (i.e. emit a `DeprecationWarning` if it was passed in explicitly – both `True` and `False` values), and remove the related code. And then, of course, update the notebooks again. :)

capellambse/svg/drawing.py
dahbar: I found that `self.obj_cache` is used in the function `get_port_transformation` on line 341. When you remove that part of the code the ports don't rotate as they should on the diagram.

Should we find a work around for this?

Wuestengecko: Ah, interesting. I didn't find that with my cursory glance. Let's keep it in, that gets out of scope otherwise.

Wuestengecko (approved):

PR #346, opened 2023-10-17
Don't crash if there are no diagrams in the model

Wuestengecko:
Just because nobody bothered to draw pretty pictures, doesn't make the whole XML invalid. Apparently diagram-less models are a thing after all.

### DIFF
--- a/capellambse/aird/__init__.py
+++ b/capellambse/aird/__init__.py
@@ -58,13 +58,8 @@ def enumerate_diagrams(
     model
         The MelodyLoader instance
     """
-    raw_views = model.xpath2(C.XP_VIEWS)
     views: list[tuple[pathlib.PurePosixPath, etree._Element, str]] = []
-    if len(raw_views) == 0:
-        raise ValueError("Invalid XML: No viewpoints found")
-
-    # Extract the views' names
-    for i in raw_views:
+    for i in model.xpath2(C.XP_VIEWS):
         viewname = helpers.xpath_fetch_unique(
             "./viewpoint", i[1], "viewpoint description"
         )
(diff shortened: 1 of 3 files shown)


capellambse/aird/__init__.py
```diff
@@ -58,13 +58,8 @@ def enumerate_diagrams(
...
         The MelodyLoader instance
     """
-    raw_views = model.xpath2(C.XP_VIEWS)
     views: list[tuple[pathlib.PurePosixPath, etree._Element, str]] = []
-    if len(raw_views) == 0:
-        raise ValueError("Invalid XML: No viewpoints found")
-
-    # Extract the views' names
-    for i in raw_views:
+    for i in model.xpath2(C.XP_VIEWS):
```
ewuerger, line 62: Has nothing to do with this PR but I still want it addressed: Just seeing `xpath2` here, naming feels off. Is `xpath2` a method we want to have public? If so maybe we rename `xpath2` to `fragment_and_elements_from_xpath`. If not wouldn't `_xpath` just fit?

ewuerger (changes requested):
Can you add a simple test for this? Just take the test model, delete all diagrams, save the model and then load it again. This should test that no `ValueError` is raised on a diagram-less model anymore.

Beside that and my concern about `xpath2`.
LGTM.

capellambse/aird/__init__.py
Wuestengecko: It's named `xpath2` because it's essentially an extended version of the `xpath` method. We use it here because we need the additional information about the fragment that each found element is in.

> Is `xpath2` a method we want to have public? If so maybe we rename xpath2 to fragment_and_elements_from_xpath. If not wouldn't `_xpath` just fit?

Honestly I don't think it _needs_ to be "public" (as in, part of the "stable public API"), but renaming it to `_xpath` would just cause confusion with `xpath` without underscore. And I do see how it might be useful for other uses as well, so keeping it public is IMO a good idea.

I'm also not a fan of overly verbose names, and 5 words is something I'd consider overly verbose for something that's really just `etree.xpath()` over the whole model. I'd just keep it like this.

<sub>Besides, this entire module will probably go away with the model-rework anyways.</sub>

Issue #339, opened 2023-09-26
Missing unconnected comments in generated images #338


alternone:
Hi,

we have large models with logical and physical architecture in Capella (5.2). During generation via capellambse we detected that for during generation for logical and physical architecture comments get lost if they are not connected to components or other architecture elements

- we expect DRepresentationDescriptor to be impacted

Attached you will find the original data with links etc. and the corrupt generated one (both with black bars to hide our super confidential content)
![image-2023-9-25_14-34-8](https://github.com/DSD-DBS/py-capellambse/assets/23116278/9482c82a-4cf4-4cfb-bb1f-c6b26da92554)

![image-2023-9-25_14-34-31](https://github.com/DSD-DBS/py-capellambse/assets/23116278/0090380c-8695-4af0-a4f5-c2405be34131)

ewuerger:
Hi,
can you elaborate what you mean by comments? From the attached pictures I guess you have comments in form of `Note`s (yellow box with folded right-upper corner). In that case I can't reproduce this behavior. 
![image](https://github.com/DSD-DBS/py-capellambse/assets/50786483/e1f354bd-b500-4f82-8fe4-25ba4bc17de9)

But I'm also missing the diagram symbol on my notes. How did you do that? Is there another diagram attached to it?

alternone:
I assume it's related to DRepresentationDescriptor on Physical Architecture. Do you use a more recent developer version?

alternone:
many thanks - I will test the results on our side!

alternone:
works as expected!

PR #347, opened 2023-10-18
Add missing representation links to diagrams

dahbar:
Fixes issue #339: A type of visual element similar to a Note called `Representation Link` was not being generated in the diagrams.

#### Screenshot of diagram with a Representation Link
d

### DIFF
--- a/capellambse/aird/__init__.py
+++ b/capellambse/aird/__init__.py
@@ -213,7 +213,9 @@ def parse_diagram(
 
 def _element_from_xml(ebd: C.ElementBuilder) -> diagram.DiagramElement:
     """Construct a single diagram element from the model XML."""
-    if ebd.data_element.get("element") is not None:
+    element = ebd.data_element.get("element")
+    tag = ebd.melodyloader[element].tag if element else None
+    if element is not None and tag != "ownedRepresentationDescriptors":
         factory = _semantic.from_xml
     else:
         factory = _visual.from_xml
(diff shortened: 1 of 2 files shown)


capellambse/aird/_visual.py
```diff
@@ -97,14 +103,21 @@ def shape_factory(ebd: c.ElementBuilder) -> diagram.Box:
         int(layout.attrib.get("width", "0")),
         int(layout.attrib.get("height", "0")),
     )
-    styleclass = ebd.data_element.attrib["type"]
+
+    if element is not None:
+        styleclass = "RepresentationLink"
+    else:
+        styleclass = "Note"
```
Wuestengecko, line 110: What's going on with this part? This used to be taken from the XML, but now it's hard-coded to "Note". Is there just no other value possible here?

Wuestengecko (commented):
I do have one question to clarify, but the rest looks really good - if this is not a big deal, it's ready to merge.

capellambse/aird/_visual.py
dahbar: Previously, the variable `styleclass` always received the value `Note` from the XML, regardless of whether it was a Note, Representation Link or Text.

The only one of those 3 that has an element attribute in the XML is Representation Link and is the only one that has slightly different styling.

I agree this is not very good and it might make sense to add something like:

```
if element is not None:
     styleclass = "RepresentationLink"
 else:
     styleclass = ebd.data_element.attrib["type"]
```
And check what other elements go into `_visual.py` and double check if everything is correct. It also makes sense to add a class for the visual element `Text` but at moment both Text and Note have the exact same styling in py-capellambse.

Wuestengecko: > I agree this is not very good and it might make sense to add something like

We don't need that. I was just wondering because I wasn't aware that "Note" was already the only possible option to begin with. :)

LGTM.

PR #352, opened 2023-10-31
Add missing attributes and default styling

ewuerger:
This is needed for [context-diagrams #55](https://github.com/DSD-DBS/capellambse-context-diagrams/pull/55) to properly work.

### DIFF
--- a/capellambse/diagram/_styleclass.py
+++ b/capellambse/diagram/_styleclass.py
@@ -30,6 +30,8 @@ def get_styleclass(obj: model.ModelObject) -> str:
 
 
 def _default(obj: model.ModelObject) -> str:
+    if isinstance(obj, model.GenericElement):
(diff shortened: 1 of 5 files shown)


capellambse/diagram/_styleclass.py
```diff
@@ -30,6 +30,8 @@ def get_styleclass(obj: model.ModelObject) -> str:
 
 
 def _default(obj: model.ModelObject) -> str:
+    if type(obj).__name__ == "GenericElement":
+        return obj.xtype.split(":")[-1]  # type: ignore[attr-defined]
```
Wuestengecko, line 34: This should use a proper `isinstance(obj, model.GenericElement)` check instead, which would also avoid the `type: ignore`.

Wuestengecko (approved):
I'll just fix this quickly myself, then it's good to merge.

PR #353, opened 2023-10-31
Rename diagram_cache to diagram_image_cache

### LINKED ISSUE #247: Rename "diagram cache" to "diagram image cache"
To make it less ambiguous what the feature is about, we should explicitly mention that it's about the graphical representations of the diagrams, i.e. images.

This includes the main entry point method `update_diagram_cache`, which constitutes a breaking API change. Therefore there will be a deprecation period.

dahbar:
Closes #247

Added a deprecation warning for `diagram_cache` and `diagram_cache_subdir` in the MelodyLoader and for the `update_diagram_cache` function.

### DIFF
--- a/capellambse/model/__init__.py
+++ b/capellambse/model/__init__.py
@@ -23,7 +23,7 @@ from lxml import etree
 import capellambse
 import capellambse.helpers
 import capellambse.pvmt
-from capellambse import _diagram_cache, filehandler, loader
+from capellambse import _diagram_image_cache, filehandler, loader
 
 from . import common, diagram  # isort:skip
 
@@ -76,7 +76,7 @@ class MelodyModel:
         capellacore.PropertyValuePkg, aslist=common.ElementList
     )
     diagrams = diagram.DiagramAccessor(
-        None, cacheattr="_MelodyModel__diagram_cache"
+        None, cacheattr="_MelodyModel__diagram_image_cache"
     )
 
     uuid = common.AttributeProperty(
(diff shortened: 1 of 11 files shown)


capellambse/model/__init__.py
```diff
@@ -238,13 +250,31 @@ def __init__(
...
             warnings.warn(
-                (
-                    "The 'jupyter_untrusted' argument is no longer needed and"
-                    " will be removed soon. Please remove it from your calls."
-                ),
+                "The 'jupyter_untrusted' argument is no longer needed and"
+                " will be removed soon. Please remove it from your calls.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
ewuerger, line 257: Has nothing to do with the `diagram_cache` rename but still good catch. Mixing these things into one commit can be annoying in the future. Keep that in mind. Even though, I forget this from time to time,...

ewuerger (approved):

capellambse/model/__init__.py
Wuestengecko: What's the "good catch"? As far as I can see, the only thing that changed was some formatting (no parenthesis around the string anymore).

dahbar: I believe these formatting changes were made by running `black --preview capellambse/model/__init__.py` and not really by me.

Wuestengecko:
We determined that we won't be doing this after all. The name "diagram cache" has already been established too well, and is also used in other places and projects (like [DSD-DBS/capella-collab-manager](https://github.com/DSD-DBS/capella-collab-manager)), changing it now would just cause unnecessary headache downstream.

PR #356, opened 2023-11-06
(Refactor) Dynamically create decorations

dahbar:
Based on a TODO from the comments:

- [x] Refactor to dynamically determine needed decorations


capellambse/svg/drawing.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, metadata: generate.DiagramMetadata):
 
         self.__drawing = drawing.Drawing(**superparams)
         self.diagram_class = metadata.class_
-        self.add_static_decorations()
+        self.deco_cache: t.List[str] = []
```
Wuestengecko, line 65: Use new-style type annotations (`list[str]` instead of `t.List[str]`).

capellambse/svg/drawing.py
```diff
@@ -389,6 +387,12 @@ def add_port(
...
 
+            if port_id not in self.deco_cache:
+                self.__drawing.defs.add(
+                    decorations.deco_factories[port_id[1:]]()
+                )
+                self.deco_cache.append(port_id)
+
             grp.add(
                 self.__drawing.use(
                     href=port_id,
```
Wuestengecko, line 398: IMO it makes more sense to add the `#` sign only where it's actually needed, instead of stripping it out where it would hurt. This way, there's this leading `#` in the `deco_cache` only for ports, but not other symbols, which seems somewhat odd to me.

Wuestengecko (commented):
Only some minor notes, the rest looks good. Great work!

PR #359, opened 2023-11-21
Optionally refresh diagrams before exporting them to the cache

Wuestengecko:
This PR adds a new flag to the `capellambse.diagram_cache` CLI, which allows to "refresh" all diagrams (i.e. synchronize them with the actual model contents) before exporting them (see aa44b4ce2dbfa328cdb5df9993d95f390b4e8272 - the other two commits merely refactor the surrounding code a little bit).

Open questions:

- [x] Do we want to enable this by default after all? (See the explanation in the docstring) ⇒ No, the Capella window is too annoying.
- [x] Do we maybe want to enable this by default in the CI template? This is a much more specific use-case, and at least doesn't have the annoying window come up. When reviewing finished documents, it's also much more apparent when an element is severely misplaced than when it's missing entirely, so it might be a good idea here to force consistency. ⇒ Yes, but that's gonna be a separate PR.

### DIFF
--- a/capellambse/_diagram_cache.py
+++ b/capellambse/_diagram_cache.py
@@ -65,6 +65,7 @@ def export(
     index: bool,
     force: t.Literal["exe", "docker", None],
     background: bool,
+    refresh: bool = False,
 ) -> None:
     if not isinstance(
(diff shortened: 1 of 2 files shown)


ewuerger (approved):
LGTM.

PR #366, opened 2023-12-01
Fix feature text alignment

ewuerger:
The feature text span x coordinate was always calculated for a centric rendering. This commit fixes it by opening the door

### DIFF
--- a/capellambse/svg/drawing.py
+++ b/capellambse/svg/drawing.py
@@ -239,9 +239,9 @@ class Drawing:
         x, y = obj.attribs["x"], obj.attribs["y"]
         w, h = obj.attribs["width"], obj.attribs["height"]
         label: LabelDict = {
-            "x": x,
+            "x": x + decorations.feature_space / 2,
             "y": y + decorations.feature_space,
-            "width": w,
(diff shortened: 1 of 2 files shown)


capellambse/svg/drawing.py
```diff
@@ -826,6 +827,11 @@ def _draw_rect_helping_lines(
         return lines
 
 
+AlignmentLiteral = t.Union[
+    t.Literal["center"], t.Literal["left"], t.Literal["right"]
+]
```
Wuestengecko, line 832:
```suggestion
AlignmentLiteral = t.Literal["center", "left", "right"]
```

capellambse/svg/helpers.py
```diff
@@ -14,6 +14,7 @@ def check_for_horizontal_overflow(
     width: int | float,
     icon_padding: int | float,
     icon_size: int | float,
+    alignment: str = "center",
```
Wuestengecko, line 17: Should this also be an `AlignmentLiteral`?

ewuerger:
> I think you're missing a part of the commit message. Which door are we opening exactly here?

Fixed that.

Issue #41, opened 2021-12-12
Undefined nature of root physical component causes KeyError on access attempt


vik378:
Physical components are supposed to have `nature` attribute, however for root physical component this attribute is undefined. Current impolementation assumes it is always defined and expects the value to match an enum. This failes with key error.

The issue can be reproduced on the test model `5_0` with the following code: `model.pa.all_components[0].nature`

![image](https://user-images.githubusercontent.com/8280844/145726320-92aa8ac7-4602-49a2-b999-81edaf0c8f9d.png)

Wuestengecko:
I'm a little unhappy with the resolution to this issue.

Obviously crashing when trying to access attributes on the root component is not an option, I absolutely agree with that. But I don't think it's a good idea to introduce an arbitrary new "dummy value" and use it as fallback. This suggests to a user who isn't familiar with the meta-model that it's actually okay to set that value on any component. Our validation logic will let that through without complaints, since the value is now registered in the underlying enum. However, the Capella GUI does not offer UNSET as a selectable option for NATURE (unlike KIND, where it does), as you can see here:

![Capella GUI showing Kind and Nature selection boxes](https://user-images.githubusercontent.com/1579756/147762514-72b4feda-b4f4-46b0-a9f2-9ed29e4e5f0a.png)

If you remove the `nature="..."` attribute from the XML using a text editor (or by assigning `UNSET` to `.nature` after d4df26303c0c946915ed6e667d17a538de150bf9), the component sort of breaks in Capella. It becomes grey in diagrams (instead of yellow or blue), and because you can't change a component's "Nature" via GUI anyways, you also can't repair it that way.

Therefore I think it would be a better idea to change the `EnumAttributeProperty`'s behavior in the "absent and no default" case from "raise an exception" to "return None", and to remove the dummy value "UNSET" from the enum of accepted values again. This would be far more graceful than raising a `KeyError(None)` into client code, yet it won't allow setting values that shouldn't be settable.

Summoning @ewuerger back for discussion.

ewuerger:
I think your arguments are well layed out and I can only agree to your proposed solution. Having an unset enum value defined on the Nature of physical components is an unnecessary deviation from the actual metamodel and happened due to me only interested in removing the KeyError raising. 
I've wrongly taken the kind enum as a template to solve this. But implementing your proposed behaviour should be fairly trivial now(Assuming the default parameter takes None). 

[...]

PR #368, opened 2023-12-06
Fix definitions of PhysicalComponent nature and kind

ewuerger:
Somehow `.nature` could return a `None` value. It should default to
`Node`.

### DIFF
--- a/capellambse/model/layers/pa.py
+++ b/capellambse/model/layers/pa.py
@@ -47,9 +47,11 @@ class PhysicalComponent(cs.Component):
 
     _xmltag = "ownedPhysicalComponents"
 
-    nature = c.EnumAttributeProperty("nature", modeltypes.Nature)
+    nature = c.EnumAttributeProperty(
+        "nature", modeltypes.PhysicalComponentNature, default="UNSET"
+    )
     kind = c.EnumAttributeProperty(
-        "kind", modeltypes.Kind, default=modeltypes.Kind.UNSET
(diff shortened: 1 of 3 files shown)


Wuestengecko (changes requested):
There was a discussion about this exact property almost exactly two years ago in #41 :)

Back then, we decided on returning `None` in case it's not set - which can only legitimately occur in the root PhysicalComponent, which act more as container for the "real" model elements instead of actually participating in the semantics.

Setting the `EnumAttributeProperty`'s `default` to "NODE" introduces a bug, as capellambse will no longer write out the required `nature="NODE"` XML attribute after it has been modified. Capella would then see it as "neither NODE nor BEHAVIOR" and display the component greyed out, rendering it pretty much useless as you can't change the nature via GUI (as discussed in the linked issue).

I took a look into the metamodel definition, and it turns out that there actually _are_ three possible enum values, and the GUI is just not showing one of them:

```xml
  <eClassifiers xsi:type="ecore:EEnum" name="PhysicalComponentNature">
    <eLiterals name="UNSET">...</eLiterals>
    <eLiterals name="BEHAVIOR" value="1" literal="BEHAVIOR">...</eLiterals>
    <eLiterals name="NODE" value="2">...</eLiterals>
  </eClassifiers>
```

So, while I was against that proposal in #41, adding a third value to use as the "UNSET" case would actually have been the correct solution. We should implement that instead - better late than never.

PR #369, opened 2023-12-22
Fix some issues in diagrams

ewuerger:
Several changes needed for the `realization_view` diagram feature in the context-diagrams project.


Wuestengecko:
This PR is a draft with no activity for a while. Did you mean to submit it for review, or is there still work to be done here?

capellambse/svg/drawing.py
```diff
@@ -294,35 +288,55 @@ def _draw_label(
             "insert": (builder.label["x"], builder.label["y"]),
             "text_anchor": builder.text_anchor,
             "dominant_baseline": "middle",
-            "style": builder.labelstyle[""],
         }
         if "class" in builder.label:
             textattrs["class_"] = builder.label["class"]
+
+        if transform := getattr(builder.labelstyle, "transform", None):
```
Wuestengecko, line 295: Maybe I'm missing a detail in this new branch, but how does this handle multi-line labels? The new code doesn't seem to call any functions related to that. Are those calls not needed here after all?

capellambse/svg/style.py
```diff
@@ -31,7 +31,7 @@ class Styling:
     def __init__(
         self, diagram_class: str | None, class_: str, prefix: str = "", **attr
     ):
-        self._diagram_class = diagram_class
+        self._diagram_class = diagram_class or ""
```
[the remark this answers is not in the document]
ewuerger: `diagram_class` can be `None`. There was a point were I didn't want to do the `or ...` stuff repetitively, which was removed again. So it is not needed anymore.

capellambse/svg/drawing.py
ewuerger: They are probably needed, but I only tested it on single line text labels.

capellambse/svg/drawing.py
```diff
@@ -25,6 +25,11 @@
 """Debug flag to render helping lines."""
 LABEL_ICON_PADDING = 2
 """Default padding between a label's icon and text."""
+NUMBER = r"?\d+(\.\d+)?"
+RE_ROTATION = re.compile(
+    rf"rotate\((?P<angle>-{NUMBER}),\s*(?P<x>-{NUMBER}),\s*(?P<y>-{NUMBER})\) "
+    f"(?P<tspan_y>-{NUMBER})"
+)
```
Wuestengecko, line 32:
```suggestion
NUMBER = r"-?\d+(\.\d+)?"
RE_ROTATION = re.compile(
    rf"rotate\((?P<angle>{NUMBER}),\s*(?P<x>{NUMBER}),\s*(?P<y>{NUMBER})\) "
    f"(?P<tspan_y>{NUMBER})"
)
```

capellambse/svg/drawing.py
Wuestengecko: We'll simply raise an error for all unsupported cases for now. Something to worry about later. :)

PR #371, opened 2024-01-16
Treat str indices to ElementList as the special case

Wuestengecko:
When determining which type of indexing is requested for an ElementList (slice, mapping-like or list-like), treat str indices (i.e. mapping-like) as the special case and integer indices as the regular case. This improves compatibility with libraries that provide their own specialized integer classes, which are not subclasses of Python's standard `int`, such as numpy's fixed-size int8/16/32/64.


ewuerger (approved):
Nice. Thanks.
LGTM.

PR #372, opened 2024-01-16
Fix-decorations

ewuerger:
Some fixes for decoration and styling of diagrams.

### DIFF
--- a/capellambse/diagram/capstyle.py
+++ b/capellambse/diagram/capstyle.py
@@ -543,19 +543,22 @@ STYLES: dict[str, dict[str, dict[str, CSSdef]]] = {
     },
     "Missions Capabilities Blank": {
         "Box.SystemComponent": {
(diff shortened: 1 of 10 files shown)


vik378 (approved):
LGTM

capellambse/diagram/capstyle.py
```diff
@@ -174,7 +174,6 @@ class in the form::
     "_CAP_Activity_Orange": RGB(247, 218, 116),
     "_CAP_Activity_Orange_min": RGB(255, 255, 197),
     "_CAP_Actor_Blue": RGB(198, 230, 255),
-    "_CAP_Actor_Blue_label": RGB(0, 0, 0),
```
Wuestengecko, line 177: This map contains all the color names known to Capella, so that the style definitions below can use the exact names used in Capella's stylesheet.

This does mean that there are dups, but I guess it is what it is.

PR #373, opened 2024-01-17
Diagram pretty-printing and saving

Wuestengecko:
Once this is released, we should also update the example snippets in the context-diagrams docs, where the svgdiagram based approach is used quite extensively.

### DIFF
--- a/capellambse/helpers.py
+++ b/capellambse/helpers.py
@@ -193,7 +193,8 @@ def load_font(fonttype: str, size: int) -> ImageFont.FreeTypeFont:
         except OSError:
             pass
 
-    with imr.open_binary("capellambse", FALLBACK_FONT) as fallback_font:
+    fontfile = imr.files(capellambse).joinpath(FALLBACK_FONT)
+    with fontfile.open("rb") as fallback_font:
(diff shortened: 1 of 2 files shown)


ewuerger (commented):
We already have a `save(pretty=True)` method on the rendered SVGDiagram. 
`There should be one-- and preferably only one --obvious way to do it.`
With this save method the user renders and then writes the rendered diagram.
Do you want to keep the save method on the SVGDiagram class?

Wuestengecko:
> Do you want to keep the save method on the SVGDiagram class?

As I discussed in the commit description, they do slightly different things, which also allows the SVGDiagram's own `save()` method to potentially support more (svg-specific) options. As concrete example, SVGDiagram.save() supports an `indent` argument, but `AbstractDiagram.save()` doesn't (it would be passed as render parameter and probably ignored). Therefore, currently both have their reasons for existing.

> There should be one-- and preferably only one --obvious way to do it.

Yeah, I'd argue that using the "svgdiagram" format instead of "svg" because that doesn't give you a plain str but a Python object which has a save method isn't exactly an "obvious" way for an unsuspecting user ;)

PR #374, opened 2024-01-18
More filepath props and add context to SVGs

huyenngn:
Capella ROS Tools uses pathlib.Path and capellambse filepath interchangeably so.suffix, .parent, .stem, .rglob filepath props were added to avoid having to check with isinstance() all the time.

Added context to diagram edges to support a feature in capella context diagrams


ewuerger (dismissed):
LGTM.

capellambse/diagram/_diagram.py
```diff
@@ -602,6 +605,7 @@ def __init__(
 
         self.labels = labels or []
         self.hidden = hidden
+        self.context = set(context) if context else set()
 
     def add_context(self, uuid: str) -> None:
         """Do nothing."""
```
Wuestengecko, line 611: This needs to be implemented to properly track context.

Wuestengecko (changes requested):

capellambse/filehandler/abc.py
```diff
@@ -251,6 +252,25 @@ def iterdir(
...
+    def suffix(self) -> str:
+        return self._path.suffix
+
+    @property
+    def stem(self) -> str:
+        return self._path.stem
+
+    @property
+    def parent(self):
+        return self._path.parent
```
Wuestengecko, line 265: `self._path.parent` is an AbstractPosixPath, which would need to be turned into an instance of `type(self)` before returning it.

Also missing return annotation: `(...) -> te.Self:`

capellambse/svg/drawing.py
```diff
@@ -666,6 +665,8 @@ def _draw_box(
             features=features_,
             children=bool(children_),
         )
+        if context_:
+            grp.add(container.Group(class_=" ".join(context_)))
```
Wuestengecko, line 669: I don't think putting the raw, undecorated IDs into the `class` attribute is a good idea. Is there a better place for this? HTML has its freely usable `data-X` attributes - does SVG maybe have something similar? If not, I'd prefer to see some prefix added to these IDs, to make it more explicit what they're meant to represent and why they're there.

Wuestengecko (changes requested):

capellambse/svg/drawing.py
ewuerger: Doesn't work. We tried some attributes, not only data-X. Maybe we didn't try it correctly? SVG-write didn't support this.

huyenngn: SVGs have data-* attributes, which we tried, as well as custom namespaces. But it wasn't supported by svgwrite.

capellambse/diagram/_diagram.py
huyenngn: Do I just copy the implementation from Box.add_context?

Wuestengecko: Well, edges can't be ports, so that part doesn't apply at least :)

I'd assume that just the `self.context.add()` call should be enough here.

capellambse/svg/drawing.py
Wuestengecko: Okay, so no data attributes - in that case, please add a short prefix like `context-` to the IDs, to make it more apparent what they are about.

PR #376, opened 2024-01-25
Don't fall back to making a random UUID

Wuestengecko:
We briefly talked about this topic before, and I vaguely recall that, in the end, we all kinda agreed on this. @ewuerger  
<sup>(I might be wrong though.)</sup>


ewuerger (approved):
I can remember and approve. LGTM.

PR #377, opened 2024-02-02
Use correct FunctionalExchange styleclass on OA

ewuerger:
The `FunctionalExchange`s in OA diagrams are not styled correctly. This is caused by a faulty/missing styleclass factory and aird factory patching the style class to `OperationalExchange`.

Additionally missing default styles were added.

### DIFF
--- a/capellambse/diagram/capstyle.py
+++ b/capellambse/diagram/capstyle.py
@@ -699,7 +699,7 @@ STYLES: dict[str, dict[str, dict[str, CSSdef]]] = {
             "marker-end": "ArrowMark",
             "stroke": COLORS["dark_gray"],
         },
-        "Edge.FunctionalExchange": {
+        "Edge.OperationalExchange": {
             "marker-end": "ArrowMark",
             "stroke": COLORS["_CAP_Activity_Border_Orange"],
         },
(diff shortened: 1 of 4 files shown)


capellambse/diagram/capstyle.py
```diff
@@ -699,7 +699,7 @@ class in the form::
             "marker-end": "ArrowMark",
             "stroke": COLORS["dark_gray"],
         },
-        "Edge.FunctionalExchange": {
+        "Edge.OperationalExchange": {
```
Wuestengecko, line 702: I'm somewhat having second thoughts about this business (our last meeting about it was a little rushed after all :) ).

We already had all these stylings defined anyways, but using the actual class "FunctionalExchange" instead of a made-up one. These styles get triggered when we're using a _diagram class_ of "Operational Entity Blank", "Operational Activity Interaction Blank", or (now also) "Operational Entity Breakdown". I recall the issue came originally somewhere from context-diagrams, but I don't know the details anymore. Specifically, as we have to use the correct diagram class anyways (otherwise "OperationalExchange" would just be a completely unstyled black line) - why can't we then just stick to "FunctionalExchange"? Do we have to make up a new class that doesn't exist in the upstream metamodel, or would it work with the real one now that OEB has the styling too?

Wuestengecko: I found it very confusing that this just suddenly started working without having to change anything in the "symbol factories", so I dug around a bit more and found that the `operational_exchange_symbol` factory (used by the "new" `OperationalExchange` class) already existed way before this PR. It was introduced before publishing this project on Github, in commit `29bdcef625f50c7fff72ca5747ed6cc86f8c2f82` from 2021-07-15, so it's by no means a new problem that we're trying to solve.

Here's the commit description from back then:

> Hack: Add OperationalExchangeSymbol to OAB
>
> The problem is that the current metamodel does not distinguish between
FunctionalExchanges on different layers. Picking the right icon only
from class-type is not working. A similiar problem exists with Functions
whose owning Component is an actor.

Funny that we're now so committed to this solution of "making up" style classes that we can't even say that it's a hack anymore. (I do still think it's somewhat hacky, but I'm fine with merging now. My curiosity is sated.)

PR #378, opened 2024-02-08
Fix SVG rendering

ewuerger:
This commit fixes too large rendered text in PDF rendered with
weasyprint. Therefore the following is done:
- Change pt to px since this is used in underlying text measuring
  function.
- Simplify icon positioning. This fixes uneven distances between icon
  and text.

Example diagram SVG with the changes to the engine:
image

### DIFF
--- a/capellambse/svg/drawing.py
+++ b/capellambse/svg/drawing.py
@@ -50,17 +50,14 @@ class Drawing:
     """The main container that stores all svg elements."""
 
     def __init__(self, metadata: generate.DiagramMetadata):
-        base_css = (
-            "shape-rendering: geometricPrecision;"
-            " font-family: 'Segoe UI';"
-            " font-size: 8pt;"
-            " cursor: pointer;"
-        )
+        base_css = "shape-rendering: geometricPrecision; cursor: pointer;"
         superparams = {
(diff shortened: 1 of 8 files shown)


Wuestengecko:
You're saying it's a breaking change - what does it break? Please add a BREAKING CHANGE footer to the commit too :)

capellambse/svg/drawing.py
```diff
@@ -911,57 +906,36 @@ def render_hbounded_lines(
...
+        builder.label["x"] - builder.icon_size - decorations.icon_padding * 2
     )
-    if (
-        icon_x < builder.label["x"] - decorations.icon_padding
-        and builder.label["class"] != "Annotation"
-    ):
-        icon_x = builder.label["x"] - decorations.icon_padding
-    return icon_x, icon_y
+    if builder.text_anchor == "middle":
+        icon_x -= lines.max_line_width / 3
```
ewuerger, line 940: Please don't ask why it is `/3` mathematically this doesn't make any sense. It should be `/2` (half) but,... the physics in this game are sth. else.

Wuestengecko (commented):
grafik

PR #382, opened 2024-02-21
fix: Repair broken sphinx references

freshavocado7:
The two new functions in the config prevent sphinx from creating documentation for __new__. The reason i chose to skip it is because the information provided by __new__ didn't add any valuable information whenever it appeared and came with broken links. 

The nitpick_ignore contains some errors which could not be resolved as of Sphinx==7.2.6. Maybe its worth looking into them in the future.


Wuestengecko (commented):
There are a few places where we need to re-wrap the text (also remember that docstrings only have 72 characters, not 79 like code).

I'll just fix these myself and merge then. Thanks for the effort :)

PR #385, opened 2024-02-26
fix: Replace references to deprecated `capella/cli` image

MoritzWeber0:
The `capella/cli` image in the [capella-dockerimages repository](https://github.com/DSD-DBS/capella-dockerimages) was deprecated in https://github.com/DSD-DBS/capella-dockerimages/pull/152. More information is available in the release notes: https://github.com/DSD-DBS/capella-dockerimages/releases/tag/v1.12.2.

While the old image is still functual, it is no longer built and therefore doesn't receive any new updates.

The `capella/cli` image was replaced with the `capella/base`. The interface is the same.

In addition, I changed one example in the documentation to point to the prebuilt Capella image from the capella-dockerimages repository.

### DIFF
--- a/capellambse/model/__init__.py
+++ b/capellambse/model/__init__.py
@@ -593,10 +593,11 @@ class MelodyModel:
         ...     "/Applications/Capella_{VERSION}.app/Contents/MacOS/capella"
         ... )
 
-        **Running a Capella CLI container**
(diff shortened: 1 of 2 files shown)


ewuerger (approved):
LGTM.

PR #386, opened 2024-02-26
Handle resources in write_tmp_project_dir

Wuestengecko:
The 'write_tmp_project_dir' method is now able to create a full workspace structure when more than one resource is involved in a model. Previously, calling this method on a model with loaded resources (i.e. library models) would result in an exception.


dominik003 (approved):
I tested it in a Gitlab pipeline and it works as expected. Thanks for the fix.

PR #389, opened 2024-03-07
Allow multiple labels on `diagram.Box`

ewuerger:
Handle multiple labels and label boxes on `diagram.Box`.


capellambse/diagram/_json_enc.py
```diff
@@ -67,8 +67,8 @@ def __encode_box(o: diagram.Box) -> object:
             "height": _intround(o.size.y),
             "context": sorted(o.context),
         }
-        if o.label is not None and not o.hidelabel:
-            jsonobj["label"] = _encode_label(o.label)
+        if o.labels is not None and not o.hidelabel:
+            jsonobj["label"] = [_encode_label(label) for label in o.labels]
```
Wuestengecko, line 71: It's weird that we encode the `labels` field into a `label` JSON property. We should unify that.

Wuestengecko:
I don't like this approach, specifically for the `Box` part. It adds even more special handling and complexity to an already very complicated part of the code base.

What about this:

- Change the `label` property of `diagram.Box` to always be a `str` (not a Box, and not even None - there's the empty string for that). This would always be used for the "text contents" of the box, and would be taken into account when auto-sizing the Box, just like it already happens.

  This already simplifies some code paths which currently have to check whether we have a Box or a str, for example:

  https://github.com/DSD-DBS/py-capellambse/blob/1fb77efaeead013b513d5647bb46c1e0c03ddb6e/capellambse/aird/_box_factories.py#L368-L373

  This would turn into a nice and simple `if box.label: box.label = f"[{box.label}]"`, which is much clearer.

- Add a separate new property for "floating labels", which is a `MutableSequence[Box]` (may be empty, but not None). These additional boxes are _not_ taken into account for auto-sizing the parent / containing Box (they're only relevant for the viewport / bounding box calculation). They will be rendered as text labels just like the singular `label`, but at a specific X/Y position instead of the automatic position inside the box. If needed, they can also have their own style overrides (or we could simply ignore them and only refer to the parent Box, to keep the code simpler and the label styling consistent).

- For edges, we can turn `label` (which can only be a Box or None, not a str) into a `labels` list with any number of Boxes inside, that part looks largely fine to me (and would even be more consistent then).

capellambse/svg/generate.py
```diff
@@ -65,8 +65,9 @@ def __init__(
         self,
         metadata: DiagramMetadata,
         objects: cabc.Sequence[ContentsDict],
+        **kwargs: dict[str, t.Any],
     ) -> None:
-        self.drawing = Drawing(metadata)
+        self.drawing = Drawing(metadata, **kwargs)
```
Wuestengecko, line 70: The constant dict-packing and -unpacking doesn't make sense, and somewhat breaks with how the same API works over in the `diagram.Diagram`. IMO we should either just pass it around as a single dict, or unpack it into regular keyword arguments.

capellambse/helpers.py
```diff
@@ -326,7 +332,7 @@ def word_wrap(text: str, width: float | int) -> list[str]:
     """Perform word wrapping for proportional fonts.
 
     Whitespace at the beginning of input lines is preserved, but other
-    whitespace is collapsed to single spaces.
+    whitespace is collapsed to single spaces. Keep words as a whole.
```
Wuestengecko, line 335: This description doesn't match the new implementation: whitespace at the beginning of lines currently is completely discarded. Please either update the description or fix the implementation. (I assume the implementation is correct, since you also updated the tests to reflect the new behavior.)

capellambse/helpers.py
```diff
@@ -341,52 +347,40 @@ def word_wrap(text: str, width: float | int) -> list[str]:
...
+                current_line = word
+        if current_line:
+            lines.append(current_line)
+        return lines
 
-        if words_count > 0:
-            output_lines.append(rejoin(words, 0, words_count))
-            if words_count < len(words):
-                input_lines.appendleft(rejoin(words, words_count, None))
+    def keep_whitespace(i: int, line: str) -> bool:
```
Wuestengecko, line 369: The `keep_whitespace` function is only called once, we should inline it.

capellambse/helpers.py
```diff
@@ -341,52 +347,40 @@ def word_wrap(text: str, width: float | int) -> list[str]:
...
+        special_chars = {"•", "-"}
+        return i == 0 or (line.lstrip() or " ")[0] in special_chars
```
Wuestengecko, line 371: We can make `special_chars` a tuple and then simply use `line.lstrip().startswith(special_chars)`.

Wuestengecko, line 371: I don't quite get why we put a space in front of the first line of input. I can kinda see the point for list items, but why unconditionally for the first line?

Wuestengecko (changes requested):
It feels like it doesn't really make sense to look at any of the label related code in aird, diagram and svg yet, until we have discussed my other comment from earlier (about splitting "labels" into "label+labels"). That would have quite large effects on some of the implementations here.

capellambse/svg/drawing.py
```diff
@@ -66,7 +71,8 @@ def __init__(self, metadata: generate.DiagramMetadata):
         self.__drawing = drawing.Drawing(**superparams)
         self.diagram_class = metadata.class_
         self.deco_cache: set[str] = set()
-        self._add_backdrop(pos=metadata.pos, size=metadata.size)
+        if not transparent_background:
+            self._add_backdrop(pos=metadata.pos, size=metadata.size)
```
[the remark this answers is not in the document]
ewuerger: It's a way to get a transparent background, which I want to use in the context-diagrams documentation. But I did disable it now b/c I also need complementary colors for the dark mode in order to make text readable again. The implementation also allows the propagation of parameters from the JSON to the drawing. So finally font-family and font-size are configurable too.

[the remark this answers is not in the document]
Wuestengecko: What do we do about pre-rendered / cached diagrams though? We don't have any way to change stuff like fonts in Capella diagrams (without completely messing up the entire layout). I suppose we could remove the backdrop rect there, but I still don't really think it's a particularly great idea to be honest. And you even found the issue yourself immediately 🙃 

> I also need complementary colors for the dark mode in order to make text readable

This is exactly the reason why the white backdrop exists. My vote goes for keeping it in and keeping our code simple.

Talking about dark mode, I have quite good experiences with simply inverting the brightness values if I don't have full control over the color palette of an image. This is a CSS filter chain I have in active use and which serves me well:

```css
filter: invert() hue-rotate(180deg) brightness(0.8)
```

You might want to experiment a bit with this or similar filters, this might completely eliminate the need for customizing the backdrop.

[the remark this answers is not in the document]
ewuerger: Look. There are many ways to achieve what I am after. I can parse the SVG and remove the backdrop rectangle... but since I know where it gets added, I wanted to make it configurable. I don't see how this makes the super simple code of initializing a Drawing more complex. Let's keep the church in the village.
When I try to render a diagram from the diagram cache (Capella made) as an `SVGDiagram` I get a RuntimeError: `Diagram not in cache:`. So no issues with changing font-family or font-size.

capellambse/svg/drawing.py
```diff
@@ -592,11 +552,15 @@ def _draw_symbol(
...
-        if label_:
-            label_["class"] = "Annotation"
+        if labels_:
+            for label in labels_:
+                label["class"] = "Annotation"
+
             self._draw_label(
                 LabelBuilder(
-                    label_,
+                    max(decorations.max_label_width, width_),
```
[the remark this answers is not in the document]
ewuerger: Good catch.

Issue #388, opened 2024-03-05
Diagram UUID inconsistent with Capella GUI


Wuestengecko:
The `uuid` reported by `capellambse.model.diagram.Diagram` objects is currently inconsistent with what the GUI reports in the expert view. The GUI shows the `uid` of the `<DRepresentationDescriptor>` object, while capellambse first follows the `repPath` link and shows the ID of the Diagram object - even though it advertises "DRepresentationDescriptor" as the `xtype`.

To avoid confusion, we should unify this by reporting the `uid` field of the descriptor, just like the Capella GUI does.

jamilraichouni:
I want to add that the GUI shows what the Java API supplies.

PR #390, opened 2024-03-12
Rework model.Diagram to use the DRepresentationDescriptor XML elements

Wuestengecko:
Use the DRepresentationDescriptor objects as primary interface to diagrams. While this does add a layer of indirection, it also solves a few issues around diagrams, allowing them to be more smoothly integrated into the model.

`model.diagram.Diagram` now uses an `etree._Element` as its `_element`. This allows it to be decorated as `@xtype_handler`, which enables `ElementList` and `model.by_uuid()` to recognize them.

Due to the representation descriptors being placed in AIRD files, which are considered not relevant for the most part, the `model.search()` method has been extended to specifically ask the `aird` module for representation descriptors if they're in the set of requested xtypes.

Note that this also changes the `uuid` reported by Diagram objects to the one on the DRepresentationDescriptor instead of the DSemanticDiagram (or similar). This matches what the Capella GUI and Java API reports, thus making interoperability easier and solving #388. However, this also means that existing diagram caches are incompatible with the new version of capellambse.


vik378 (commented):
looks really good, why don't we merge & release?

PR #394, opened 2024-03-26
Declarative: Rename modify to set

huyenngn:
~~sync set should be called sync modify~~ modify should be called set. it's weird that it has to be:
```
- parent: !uuid blahblah
  modify:
    description: "hey"
```
for parent, but for find it's:
```
- find:
    name: blah
  set:
    description: "hey"
```
even though they both do the same thing.

it should instead be:
```
- parent: !uuid blahblah
  set:
    description: "hey"
```

### DIFF
--- a/capellambse/decl.py
+++ b/capellambse/decl.py
@@ -219,6 +219,17 @@ def _operate_modify(
     parent: capellambse.ModelObject,
     modifications: dict[str, t.Any],
 ) -> cabc.Generator[_OperatorResult, t.Any, None]:
+    import warnings
+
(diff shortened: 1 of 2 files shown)


Wuestengecko:
> it's weird that it has to be "modify" for parent, but for sync it has to be "set"

That's actually a good point, I didn't realize this during #387. But I do think that "set" makes more sense specifically in the sync context (because we might not have anything to modify yet), so we should rather rename "modify" to "set" instead (and keep "modify" as alias).

PR #396, opened 2024-04-02
feat: Implement replacing model objects for RoleTagAccessors

huyenngn:
i need this for syncing class properties to work


capellambse/model/common/accessors.py
```diff
@@ -1769,9 +1769,7 @@ def __set__(
                     "Moving model objects is not supported yet"
                 )
             if self.__get__(obj) is not None:
-                raise NotImplementedError(
-                    "Replacing model objects is not supported yet"
-                )
+                obj._element.remove(obj._element.find(self.role_tag))
```
Wuestengecko, line 1772: We first need to deregister the XML subtree from the ID cache before actually removing anything.

Wuestengecko, line 1772: The call to `self.__get__()` will give us the constructed proxy for the existing element. We can use its return to avoid the extra `find()`.

PR #406, opened 2024-04-17
Use Accessor for OperationalActivity inputs/outputs

ewuerger:


capellambse/model/layers/oa.py
```diff
@@ -31,15 +31,15 @@ class OperationalActivity(fa.AbstractFunction):
...
 
-    @property
-    def inputs(self) -> c.ElementList[fa.FunctionalExchange]:
-        return self._model.oa.all_activity_exchanges.by_target(self)
+    inputs = c.ReferenceSearchingAccessor(
+        fa.FunctionalExchange, "target", aslist=c.MixedElementList
+    )
+    outputs = c.ReferenceSearchingAccessor(
+        fa.FunctionalExchange, "source", aslist=c.MixedElementList
+    )
```
Wuestengecko, line 40: We're only looking at `FunctionalExchange` instances, so these should use the normal `ElementList`.

PR #412, opened 2024-04-24
Avoid element duplication in decl sync

huyenngn:


capellambse/decl.py
```diff
@@ -297,7 +297,8 @@ def _operate_sync(
                 yield from _create_complex_objects(
                     promises, parent, attr, [newobj_props]
                 )
-                yield from _operate_sync(promises, parent, {attr: [obj]})
+                if "sync" in obj:
+                    yield from _operate_sync(promises, parent, {attr: [obj]})
```
Wuestengecko, line 301: I'm curious how this duplication actually gets triggered. Do you mind adding a small test case for it, also to make sure that it won't happen again?

PR #413, opened 2024-04-24
Ignore labels with transforms from bounds calculation

ewuerger:
Text transforms are used for translations and rotations. Right now we only use rotations, such that positions can be very large since they are transformed anyways. I propose to trust the user with the calculated points instead of checking them during our calculations.

**_This is needed to fix unnecessary whitespace in realization view diagrams in the [context-diagrams extension #90](https://github.com/DSD-DBS/capellambse-context-diagrams/pull/90)._**

### DIFF
--- a/capellambse/diagram/_diagram.py
+++ b/capellambse/diagram/_diagram.py
@@ -456,15 +456,25 @@ class Box:
 
     @property
     def bounds(self) -> Box:
-        """Calculate the bounding box of this Box."""
+        """Calculate the bounding box of this Box.
+
+        Notes
+        -----
+        Labels with a `text_transform` in its `styleoverrides` are
+        ignored during bounds calculation.
+        """
         minx, miny = self.pos
         maxx, maxy = self.pos + self.size
+        if self.styleoverrides.get("text_transform") is not None:
+            return Box((minx, miny), (maxx - minx, maxy - miny))
+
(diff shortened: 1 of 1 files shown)


Wuestengecko (commented):
Box labels shouldn't have any style overrides at all. They're just simply ignored during SVG conversion, and so it should be an error (or I guess "undefined behavior") in any other case.

https://github.com/DSD-DBS/py-capellambse/blob/799c529cc0824be59bac8ff5c5ca814fada2bc78/capellambse/diagram/_json_enc.py#L121-L130

Unless you're suggesting to change this as well? Then I'd say it makes sense to do that in the same PR.

ewuerger:
> Box labels shouldn't have any style overrides at all. They're just simply ignored during SVG conversion, and so it should be an error (or I guess "undefined behavior") in any other case.

They are not ignored because they come from the style of the Box:
https://github.com/DSD-DBS/py-capellambse/blob/799c529cc0824be59bac8ff5c5ca814fada2bc78/capellambse/diagram/_json_enc.py#L54-L89

This is done by `text_style`:
https://github.com/DSD-DBS/capellambse-context-diagrams/blob/main/capellambse_context_diagrams/context.py#L538-L561

Then this style override is resolved in `draw_object`.

PR #424, opened 2024-05-21
Icons

Wuestengecko:
This PR exposes the object icons used in the SVG renderer as standalone icons, primarily through the new (public) `_get_icon()` instance method on model objects.


capellambse/diagram/_icons.py
```diff
@@ -0,0 +1,1240 @@
...
+        center=center,
+        r=r,
+        focal=focal,
+        gradientUnits="userSpaceOnUse",
+        gradientTransform=(
+            f"matrix({' '.join(map(str, matrix))})"
+            if matrix is not None
+            else None
+        ),
+    )
```
ewuerger, line 1200:
```suggestion
    params = {
        "id_": id_,
        "center": center,
        "r": r,
        "focal": focal,
        "gradientUnits": "userSpaceOnUse",
    }
    if matrix is not None:
        params["gradientTransform"] = f"matrix({' '.join(map(str, matrix))})"

    grad = gradients.RadialGradient(**params)
```

Somehow this fails if matrix is `None`.

tests/test_svg.py
```diff
@@ -141,53 +141,6 @@ def test_diagram_decorations(
...
-            attr: (0, 0, 0),
-        }
-        with pytest.raises(ValueError):
-            symbols._make_lgradient(**params)
-
-    def test_making_linear_gradient_with_translate(self):
-        gradient = symbols._make_lgradient("test", translate=(0, 0))
-        assert gradient.attribs.get("gradientTransform") == "translate(0 0)"
-
-
```
ewuerger, line 190: I don't like that we remove testing of the icon factory and don't add it again. The Factory moved now into the diagram._icons module. This isn't exposed but still we should test it?!

ewuerger (changes requested):
I tested rendering of ContextDiagrams from the documentation and against a large internal model. It only failed once, a fix has been suggested.

I really emphasize on not reducing coverage with tests. If this is super urgent I'm fine with opening an issue to regain the lost tests for the factory, else I suggest to get them implemented before merging this in.

LGTM then.

tests/test_svg.py
Wuestengecko: I removed these tests because they weren't really doing anything useful:

- 1 and 2 spot-check the factory registration. Basically useless, considering the number of keys that also should exist but aren't tested. (They really should have tested the decorator behind the factory registration, but that is almost trivial and IMO not worth testing either - you can't even say the regex there is complicated.)
- 3 checks that specific things get logged. Why?
- 4 and 5 were really testing svgwrite and not our code
 
If you have some ideas for tests that make sense in this context, or you disagree on what is worth testing here, let me know tho :)

capellambse/diagram/_icons.py
Wuestengecko: I might have found a better one, can you please retest against https://github.com/DSD-DBS/py-capellambse/pull/424/commits/d093c529742e4edaab4ab02cf59a9b058b9c230f?

Wuestengecko: We tested it together, and it works fine.

PR #428, opened 2024-05-28
Modeling rules and validation

freshavocado7:
Based on the modeling rules maintained in our internal Polarion and the old PR #270 that outlined the work for me, I have implemented the modeling rules in py-capellambse.

To generate the metrics report, run the following command:

    python -m capellambse.extensions.validation -m [model] -o report.html

This command will create a report.html file containing all the applied modeling rules and the metrics report.

### DIFF
--- a/capellambse/extensions/validation/__main__.py
+++ b/capellambse/extensions/validation/__main__.py
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import logging
+import typing as t
+
+import click
+import jinja2
+
+import capellambse
+
+
+@click.command()
+@click.option("-m", "--model", required=True, type=capellambse.ModelCLI())
+@click.option(
+    "-o",
+    "--output",
+    type=click.File("w", atomic=True),
+    required=True,
+    help="Output file to render the template into",
+)
+@click.option("-t", "--template", help="An optional custom template to render")
+def _main(
+    model: capellambse.MelodyModel,
+    template: str | None,
+    output: t.IO[str],
+) -> None:
+    logging.basicConfig()
+
+    loader: jinja2.BaseLoader
+    if template is None:
+        loader = jinja2.PackageLoader("capellambse", "extensions/validation")
+        template = "report-template.html.jinja"
+    else:
+        loader = jinja2.FileSystemLoader(".")
+    env = jinja2.Environment(loader=loader)
+
+    with output:
+        env.get_template(template).stream(
+            model=model,
+            results=model.validation.validate(),
+        ).dump(output)
+
+
+if __name__ == "__main__":
+    _main()
--- a/capellambse/extensions/validation/_validate.py
+++ b/capellambse/extensions/validation/_validate.py
@@ -1,0 +1,465 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+__all__ = [
+    "Category",
+    "ElementValidation",
+    "ModelValidation",
+    "RealType",
+    "Result",
+    "Results",
+    "Rule",
+    "Rules",
+    "Validation",
+    "VirtualType",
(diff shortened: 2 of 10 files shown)


Wuestengecko:
We need to clarify the relation between this and #270. I thought that #270 was to be split up, but now this completely new PR also contains large parts of it squashed in. Assuming this gets merged, what would happen to #270? Would it be abandoned? Rebased? Split out into its own extension repo?

capellambse/extensions/validation/_validate.py
```diff
@@ -0,0 +1,483 @@
...
+    category: Category,
+    types: (
+        str
+        | VirtualType[_T]
+        | type[_T]
+        | cabc.Iterable[str | VirtualType[_T] | type[_T]]
+    ),
+    id: str,
+    name: str,
+    rationale: str,
```
Wuestengecko, line 344: The `rationale` can become quite long, and is currently always passed using implicitly concatenated string fragments. For ergonomic reasons, IMO it would make sense to allow passing a triple-quoted multi-line string (just like docstrings), which gets cleaned up with something like `inspect.cleandoc`.

capellambse/extensions/validation/_validate.py
```diff
@@ -0,0 +1,483 @@
...
+        type_names = [types]
+    else:
+        type_names = []
+        for i in types:
+            if isinstance(i, str):
+                type_names.append(i)
+            elif isinstance(i, (RealType, VirtualType)):
+                type_names.append(i.name)
+            else:
+                type_names.append(i.__name__)
```
Wuestengecko, line 382: `types` should end up being a tuple, to prevent it accidentally being modified later.

capellambse/extensions/validation/_validate.py
```diff
@@ -0,0 +1,483 @@
...
+        return typeobj.search(self._model)
+
+
+class ElementValidation(Validation):
+    """Provides access to the model's validation rules and results."""
+
+    @property
+    def rules(self) -> Rules:
+        """Return all registered validation rules for this ModelObject."""
+        return self._model.validation.rules.by_type(type(self._element))
```
Wuestengecko, line 473:
```suggestion
    @property
    def rules(self) -> Rules:
        """Return all registered validation rules for this ModelObject."""
        return VALIDATION_RULES.by_type(type(self._element))
```

capellambse/extensions/validation/_validate.py
```diff
@@ -0,0 +1,483 @@
...
+        return VALIDATION_RULES
+
+    @property
+    def results(self) -> Results:
+        """Return all stored validation rule results."""
+        key = f"{__name__}.RESULTS"
+        if key not in self._model.__dict__:
+            self._model.__dict__[key] = Results()
+            self.validate()
+        return self._model.__dict__[key]
```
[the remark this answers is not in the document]
Wuestengecko: I just realized that my first point (uninitialized Results object) wouldn't actually happen, because on first access validation does happen; however, this also means that an explicit call to `validate()` performs the full model validation twice. This sort of logic error obviously wouldn't have been possible either.

capellambse/extensions/validation/__main__.py
```diff
@@ -0,0 +1,66 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG
+# SPDX-License-Identifier: Apache-2.0
+
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capellambse contributors
+# SPDX-License-Identifier: Apache-2.0
```
Wuestengecko, line 5: Double header

capellambse/extensions/validation/_validate.py
```diff
@@ -0,0 +1,483 @@
...
+    def __call__(self, obj: _T) -> bool:
+        return self.validator(obj)
+
+    def __hash__(self) -> int:
+        return hash(self.id)
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, str):
+            return self.id == other
+        return hash(self) == hash(other)
```
Wuestengecko, line 194: This is just plain wrong. `hash(a) == hash(b)` **does not** imply that `a == b`, only the other way around.

PR #433, opened 2024-07-01
ci: Only trigger pypi push on release tag

ewuerger:
The pipeline on master is unsuccessful because only non-dev versions can be uploaded to PyPI. This fixes the situation by only triggering the pypi stage when a release tag was created.

### DIFF
--- a/.github/workflows/code-qa.yml
+++ b/.github/workflows/code-qa.yml
@@ -85,7 +85,7 @@ jobs:
     name: Publish to PyPI
     runs-on: ubuntu-latest
     needs: [build, test]
-    if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/master'
+    if: startsWith(github.ref, 'refs/tags/v')
     environment:
       name: pypi
       url: https://pypi.org/project/capellambse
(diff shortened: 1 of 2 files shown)


Wuestengecko:
The pipeline on master is unsuccessful because I force pushed to fix the commit message of the latest merge, and PyPI doesn't let you reupload the same version twice. :)

```
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
         File already exists. See https://pypi.org/help/#file-name-reuse for
         more information.
```

Regardless, it might not be necessary to upload all dev versions to PyPI after all. Space there is limited, and I'm not sure anymore why we did that in the first place (probably something to do with misconfigured firewalls or something). Do you have any idea? If not, I'd maybe actually merge this and see what breaks 🙃

Wuestengecko:
> Maybe an **and** (&&?) is correct here?

Definitely not, because the ref either is `refs/heads/master` or starts with `refs/tags/v`, but never both at the same time ;)

Publishing pre-releases was a deliberate choice (81cccd844efc3841d56b328999c85bcd07a00271), but the commit doesn't have a reason either. Let's just go through with this PR. The way to install pre-releases henceforth shall be to just use the git master directly like for any other repo.

But first we'll have to somehow deal with the Sphinx failure.

PR #437, opened 2024-07-10
Git: Check out the work tree

### LINKED ISSUE #427: Committing to Git sometimes fails with "object is unavailable"
Sometimes, when trying to save a Model back to a Git repository, the save fails with a rather cryptic error message, similar to:

```
ERROR:capellambse.filehandler.git.git:fatal: entry '<filename>' object <hash> is unavailable
ERROR:capellambse.filehandler.git:Exit status: 128
Traceback (most recent call last):
[...snip...]
  File "/opt/.venv/lib/python3.11/site-packages/capellambse/filehandler/git.py", line 295, in __exit__
    tree = self.__update_tree(
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/.venv/lib/python3.11/site-packages/capellambse/filehandler/git.py", line 443, in __update_tree
    self.__handler._git(
  File "/opt/.venv/lib/python3.11/site-packages/capellambse/filehandler/git.py", line 944, in __git_nolock
    proc = subprocess.run(
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'mktree', '-z']' returned non-zero exit status 128.
```

This probably happens because not all files are downloaded every time, but for hashing a tree object, all blobs in the tree must be locally available (and won't be automatically fetched).

Wuestengecko:
This commit changes the Git file handler to check out the work tree as real files. This simplifies some operations drastically, as it allows us to rely more on the porcelain layer and specifically Git's index. By having all files of the given commit checked out, it also forces all blobs to be made available locally, and therefore resolves #427.

As part of this change, the base directory for the work trees was moved from the system temporary directory to the user cache directory. Modern Linux systems have the former reside in a ramdisk, which is not a good place to store (potentially huge) model files.

### DIFF
--- a/capellambse/filehandler/git.py
+++ b/capellambse/filehandler/git.py
@@ -1,21 +1,11 @@
 # SPDX-FileCopyrightText: Copyright DB InfraGO AG
 # SPDX-License-Identifier: Apache-2.0
 
-# pylint: disable=abstract-method, useless-suppression
-# For some reason, pylint in Github CI didn't get the memo that these aren't
-# actually abstract methods. Other pylint installations seem to agree that
-# implementing these methods isn't necessary. So we just ignore the warning
-# about that here.
-# TODO Revisit this decision some time in the future
-
 from __future__ import annotations
 
 import collections.abc as cabc
 import errno
-import functools
 import hashlib
-import io
-import itertools
 import logging
 import os
 import pathlib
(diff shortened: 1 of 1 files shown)


Wuestengecko:
Go home commitlint, you're drunk.

Issue #78, opened 2022-04-08
Drawings of Class diagrams


henrik997:
There are a few drawing issues in Class diagrams:

1. We disregard the aggregation kind of associations, in the following diagram the source ending of the edge should be an _Diamond-Marker_. 

![image](https://user-images.githubusercontent.com/44259278/162394948-bd4defd4-6b2d-442f-b538-5ea89b1914e1.png)

2. Specific box types are lacking as e.g. `Box.NumericType` `Box.StringType` for ` Class Diagram Blank.` They have to be added in _capstyle.py_ 

![image](https://user-images.githubusercontent.com/44259278/162396347-992e825b-81d3-48e2-9f46-0dccb8a1d234.png)

3. Symbols of properties are lacking. 

![image](https://user-images.githubusercontent.com/44259278/162397958-47c8632e-d5c3-4e54-bae7-19972e83b88b.png)

Actually:

![image](https://user-images.githubusercontent.com/44259278/162398045-c03548ec-2fc8-4641-9dce-e92bd9f0bec6.png)

4. Multiplicities are lacking.

![image](https://user-images.githubusercontent.com/44259278/162398246-301a32a8-7505-417c-aed4-5bde3a11caca.png)

Actually: 

![image](https://user-images.githubusercontent.com/44259278/162398139-eb8ffc4f-41c3-41e6-b90f-9ccde946a3a8.png)

5. ~~The Fine-ArrowMark looks a bit overdimensioned.~~

![image](https://user-images.githubusercontent.com/44259278/162399673-b457494c-2707-4cd8-b8a5-3c7e8f0ef250.png)
 
Solved in branch _code-generation_.

PR #449, opened 2024-08-20
fix: Add icons and multiplicity to class diagrams

huyenngn:
Partially solves #78

solved:
- Added class property, enum and enum literal icons
- Added multiplicity to class diagram edges

not solved:
- didn't add icons for Box.NumericType and Box.StringType because why would you need that
- aggregation kind of associations still don't have the Diamond-Marker when kind attribute isn't in the XML. not possible to implement atp

### DIFF
--- a/capellambse/svg/drawing.py
+++ b/capellambse/svg/drawing.py
@@ -258,29 +258,30 @@ class Drawing:
         x = obj.attribs["x"]
         y = y or obj.attribs["y"] + decorations.feature_space
         w, h = obj.attribs["width"], obj.attribs["height"]
-        label: LabelDict = {
-            "x": x + decorations.feature_space / 2,
-            "y": y,
-            "width": w - decorations.feature_space / 2,
-            "height": h - decorations.feature_space,
-            "class": "Features",
-            "text": "\n".join(
-                chelpers.flatten_html_string(feat) for feat in features
-            ),
-        }
-        self._draw_label(
-            LabelBuilder(
-                w,
-                h,
-                [label],
-                group,
-                class_=class_,
-                labelstyle=labelstyle,
-                y_margin=5,
-                icon=False,
-                alignment="left",
+        for feat in features:
+            label: LabelDict = {
+                "x": x + decorations.feature_space / 2,
+                "y": y,
+                "width": w - decorations.feature_space / 2,
+                "height": h - decorations.feature_space,
+                "class": "Features",
+                "text": chelpers.flatten_html_string(feat),
+            }
+            lines = self._draw_label(
+                LabelBuilder(
+                    w,
+                    h,
+                    [label],
+                    group,
+                    class_=f"{class_}Feature",
+                    labelstyle=labelstyle,
+                    y_margin=5,
+                    icon=True,
+                    alignment="left",
+                )
             )
-        )
+            if lines:
+                y += lines.text_height
 
     def _draw_label(self, builder: LabelBuilder) -> LinesData | None:
         """Draw label text on given object and return the label's group."""
@@ -974,7 +975,7 @@ def get_label_position(
         label_height = chelpers.extent_func(label["text"])[1]
         dominant_baseline_adjust = label_height / 2
         if builder.text_anchor == "start":
-            label["x"] += lines.margin + icon_size
+            label["x"] += lines.margin + icon_size / 2
         else:
             label["x"] += (label["width"] + icon_size) / 2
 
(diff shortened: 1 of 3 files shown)


capellambse/svg/drawing.py
```diff
@@ -970,11 +971,11 @@ def get_label_position(
...
     assert lines
 
+    label_height = chelpers.extent_func(builder.labels[0]["text"])[1]
+    dominant_baseline_adjust = label_height / 2
     for label in builder.labels:
-        label_height = chelpers.extent_func(label["text"])[1]
-        dominant_baseline_adjust = label_height / 2
         if builder.text_anchor == "start":
-            label["x"] += lines.margin + icon_size
+            label["x"] += lines.margin + icon_size / 2
```
ewuerger, line 978: This possibly has an effect on other diagrams. As labels are one of the biggest pain points for the diagram rendering engine, extra caution is needed here. The changes done here do:
* fixate the label_height from the first line of the label (from dynamic to static line_height)
* Reduce the distance of the icon to the label itself, by half of the icon_size (which is normally 25x25 therefore 12.5)

Did you check if other diagram type rendering is still ok?

capellambse/svg/drawing.py
```diff
@@ -258,29 +258,30 @@ def _draw_feature_text(
...
+                    w,
+                    h,
+                    [label],
+                    group,
+                    class_=f"{class_}Feature",
+                    labelstyle=labelstyle,
+                    y_margin=5,
+                    icon=True,
+                    alignment="left",
+                )
```
ewuerger, line 281: We call `_draw_label` now for each feature to get an icon for each I guess?

ewuerger (changes requested):
Just my thoughts about the PR. The class diagrams look better now.

Please also check this @Wuestengecko

capellambse/svg/drawing.py
huyenngn: 1. Since we calculate `label_height` with `chelpers.extent_func` instead of `chelpers.get_text_extent` and all labels of a `LabelBuilder` object have the same font size, shouldn't `label_height` always equal to the same value? Maybe we should use `chelpers.get_text_extent` instead then to account for potential line wrapping.
2. This doesn't reduce the distance of the icon to the label, as icon position is calculated relative to label position, but the distance of the box's left border to the label.
I don't know the reason but even when leaving `label["x"]` as is there is a giant gap between the label and the box's border.
I haven't checked with other diagram types as I don't really know any other diagram types where `text_anchor` defaults to "start".

ewuerger: > 1. Since we calculate `label_height` with `chelpers.extent_func` instead of `chelpers.get_text_extent` and all labels of a `LabelBuilder` object have the same font size, shouldn't `label_height` always equal to the same value? Maybe we should use `chelpers.get_text_extent` instead then to account for potential line wrapping.

No. The `label_height` is determined by 3 different values: Baseline, upper and lower extensions (there are dedicated terms for this, which I can't remember). So for example:

`acore` is a word which has only the baseline in its height, `bookstore` has baseline and the upper extension through `b`, `k` and `t`. `Any` has a lower extension through `y`, an upper by `A` and baseline as always. So the `line_height` should be determined by iterating through all letters and picking up the maximum height of each glyph... We are using the font measure functionality of Pillow with the `getsize` method on FreeFontType. [Looking into the docs](https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#font-size-and-offset-methods),... this way isn't supported anymore and we should use a different method: `getbbox` or `getlength`. 

>     2. This doesn't reduce the distance of the icon to the label, as icon position is calculated relative to label position, but the distance of the box's left border to the label.
>        I don't know the reason but even when leaving `label["x"]` as is there is a giant gap between the label and the box's border.
>        I haven't checked with other diagram types as I don't really know any other diagram types where `text_anchor` defaults to "start".

I have no clue too. I wrote the logic, but I can't remember without reading the lines of code.

Wuestengecko: As Ernst described, there can be very subtle differences in height between the different labels due to the letters used. So it makes sense to look at each label individually, instead of taking the first label for all the others as well.

I've also tested across a number of diagrams from one of our larger internal models, and there have indeed been differences in height (although expectedly very small, during the spot-check I didn't see anything larger than 3 pixels in total). Still, the "right thing" here IMO is to look at all the labels as individual components.

Other than this, the PR looks good to me, so I'm just going to change this bit myself and merge.

PR #450, opened 2024-09-02
fix: Fix snapping of oblique edges

### LINKED ISSUE #448: Edge labels are mispositioned
edge routing isn't always correct either. sometimes there is no knick when there should be a knick in the edge

huyenngn:
Fixes #448


capellambse/aird/_edge_factories.py
```diff
@@ -310,9 +311,18 @@ def snap_oblique(
...
 ) -> None:
     """Snap ``points``' end to ``target`` in a straight line."""
-    points[i] = target.vector_snap(
-        points[i], source=points[next_i], style=diagram.RoutingStyle.OBLIQUE
+    source = points[next_i]
+    new_point = target.vector_snap(
+        points[i], style=diagram.RoutingStyle.OBLIQUE, source=source
     )
+    delta = (points[i] - source).angleto(new_point - source)
+    if (math.ceil(delta) != 0) and (math.floor(delta) != 0):
```
Wuestengecko, line 319: That's an odd way of saying "if abs(delta) >= 1".

Wuestengecko (dismissed):
LGTM.

Issue #426, opened 2024-05-23
Enhancement of State / Mode objects


vik378:
In Capella itself you can do a lot more with states and modes at the moment, lets fix that. For State / Mode objects we should add the following attributes:

- `.related_transitions` - list of transitions that may lead to and from this state
- `.incoming_transitions` - list of transitions that lead to this state
- `.outgoing_transitions` - list of transitions that lead out of this state
- `.owner` - a structural element (Entity, Component, Class, etc) that the parent state machine belongs to
- `.related_states` - list of states that the owner may transition to or from this state
- `.source_states` - list of states from which the system could transition to this state
- `.target_states` - list of states to which the system could transition from this state
- `.entry` - list of activities or functions that are performed on entry into state
- `.do_activity` - list of activities or functions that may be performed while in state
- `.exit` - list of activities or functions that are performed before leaving the state

we should also rename `.states` to `.states_and_modes` and add a separate `.states` that isnt mixed with modes

Adding the above should also enable introduction of a meaningful context diagram for a state.

PR #455, opened 2024-09-29
Enhance *State/Mode classes with incoming and outgoing StateTransition accessors

chgio:
Hello py-capellambse team, and thank you very much for this great project and especially for the amazing contributor experience!

I'm interested in working with `StateMachine`s and the relations within and around them, so I noticed #426 which I'll now stop spamming with mentions. To start from the related `StateTransition`s, this PR adds:
- `incoming_transitions`
- `outgoing_transitions`

Please check the commits for the detailed lists of classes impacted, which can also be rolled back to just `State`.

The PR also adds test cases for both of these new accessors on `State`, and for `InitialPseudoState.outgoing_transitions`; I couldn't find a `StateMachine` with a `FinalState` to write the test case for `FinalState.incoming_transitions` but it should be easy to add with the right UUIDs.

Thanks in advance!

### DIFF
--- a/capellambse/metamodel/capellacommon.py
+++ b/capellambse/metamodel/capellacommon.py
@@ -49,6 +49,9 @@ class State(AbstractStateMode):
         m.ModelElement, "exit", aslist=m.MixedElementList
     )
 
+    incoming_transitions = m.Accessor
+    outgoing_transitions = m.Accessor
+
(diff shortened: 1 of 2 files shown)


Wuestengecko (approved):
Neat, thanks!

PR #459, opened 2024-10-15
Move PhysicalLink and -Port styles to `__GLOBAL__`

ewuerger:
Apparantly `PhysicalLink` and `PP` (PhysicalPorts) can exist in the System Analysis and Logical Architecture Layer. But there is no styling in the registry in capstyle.py yet. This commit fixes it.

### DIFF
--- a/capellambse/diagram/capstyle.py
+++ b/capellambse/diagram/capstyle.py
@@ -281,6 +281,10 @@ STYLES: dict[str, dict[str, dict[str, CSSdef]]] = {
             "stroke": RGB(255, 204, 102),
             "text_fill": RGB(0, 0, 0),
         },
+        "Box.PP": {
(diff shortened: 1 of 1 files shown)


capellambse/diagram/capstyle.py
```diff
@@ -320,6 +324,11 @@ class in the form::
             "stroke-dasharray": "5",
             "text_fill": COLORS["dark_purple"],
         },
+        "Edge.PhysicalLink": {
+            "stroke": COLORS["red"],
+            "stroke-width": 2,
+            "text_fill": COLORS["red"],
+        },
```
Wuestengecko, line 331: Please keep this dict sorted alphabetically. That's the only sliver of sanity we can cling to.

PR #463, opened 2024-10-17
Add component and functional exchange categories

### LINKED ISSUE #462: Feature Request: Support for Component Exchange Categories
In SA, LA and PA layers, the Component Exchanges can be assigned one or more Component Exchange Categories.
Currently, the access to the categories has no API equivalent in the capellambse package.

Wuestengecko:
Resolves #462

### DIFF
--- a/capellambse/metamodel/la.py
+++ b/capellambse/metamodel/la.py
@@ -33,6 +33,9 @@ class LogicalFunctionPkg(m.ModelElement):
     )
 
     packages: m.Accessor
+    categories = m.DirectProxyAccessor(
+        fa.ExchangeCategory, aslist=m.ElementList
+    )
 
 
 @m.xtype_handler(None)
@@ -71,6 +74,9 @@ class LogicalComponentPkg(m.ModelElement):
     )
 
     packages: m.Accessor
+    exchange_categories = m.DirectProxyAccessor(
+        fa.ComponentExchangeCategory, aslist=m.ElementList
+    )
 
 
 @m.xtype_handler(None)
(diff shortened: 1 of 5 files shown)


capellambse/metamodel/la.py
```diff
@@ -33,6 +33,9 @@ class LogicalFunctionPkg(m.ModelElement):
     )
 
     packages: m.Accessor
+    categories = m.DirectProxyAccessor(
```
podolsir, line 36: Shouldn't this be exchange categories in analogy to LogicalComponentPkg?
Currently there seems to be this "visual" imbalance between
```model.la.component_package.exchange_categories # -> produces list of _contained_ ECs```
vs.
```model.la.function_package.categories # -> produces list of _contained_ ECs```
vs.
```model.la.component_exchanges[0].categories -> produces list of _allocated_ ECs for this exchange```

Wuestengecko: That's actually a good point, especially the "allocated vs. contained" bit. Currently we don't distinguish between these via attribute names except for a few rare cases where it's necessary (for example because both exist on the same class).

The Capella metamodel usually solves this by prefixing (almost all) containment relations with "owned", e.g. "ownedCategories" vs. "categories", but at some point in the past we decided for capellambse that having this prefix basically everywhere would just end up making the library more annoying to work with. On the other hand, this now has led to some cases where it's not really clear where the objects are actually stored.

For the time being, I would simply keep it like this. I'm currently working on overhauling a lot of the internals for v0.7 over in #456, which also serves as an opportunity to extend the metamodel with lots of the parts we're currently missing - this has already uncovered a few other shortcomings of the current naming scheme, so we'll have to come up with something better soon anyways.

PR #464, opened 2024-10-22
feat: Add support for embedded images

MoritzWeber0:
Currently, embedded images are not handled properly. An embedded image link contains the name of the Capella project as first element of the path in the `src` attribute, but the path doesn't exist on the file system.

To resolve the issue, we remove the project name as prefix.

Open issues: 
- If the description is not modified but written back, there is an unintended diff. We have to make sure that the project name is added again in the `_to_xml` method. However, I don't have a stable solution how to identify embedded image links yet.
- Tests should be added to cover embedded images.
- It doesn't work if the image is referenced from another project.

Alternative solution could be to provide a description postprocessing helper function, which is used for the HTML rendering. This would solve the issue with broken image references in Jupyter.

### DIFF
--- a/capellambse/model/_pods.py
+++ b/capellambse/model/_pods.py
@@ -23,7 +23,9 @@ import typing as t
 
 import markupsafe
 import typing_extensions as te
+from lxml import etree
 
+import capellambse
 from capellambse import helpers
 
 from . import E, U
@@ -77,7 +79,7 @@ class BasePOD(t.Generic[U]):
         assert data is None or isinstance(data, str)
         if data is None:
             return self.default
-        return self._from_xml(data)
+        return self._from_xml(data, obj._model)
 
     def __set__(self, obj: t.Any, value: U | None) -> None:
         if not self.writable and self.attribute in obj._element.attrib:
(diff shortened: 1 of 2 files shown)


capellambse/model/_pods.py
```diff
@@ -101,7 +103,9 @@ def __set_name__(self, owner: type[t.Any], name: str) -> None:
         self.__objclass__ = owner
 
     @abc.abstractmethod
-    def _from_xml(self, value: str, /) -> U: ...
+    def _from_xml(
+        self, value: str, model: capellambse.MelodyModel, /
+    ) -> U: ...
```
Wuestengecko, line 108: Minor thing: Please change the order to be `self, model, value` instead of `self, value, model`, for more consistency overall. <sub>(Almost)</sub> All functions that require a model expect it as first non-self argument.

It's an API break anyways, as old methods only expect a single argument.

capellambse/model/_pods.py
```diff
@@ -151,14 +156,38 @@ def __init__(self, attribute: str, /, *, writable: bool = True) -> None:
...
-        return markupsafe.Markup(value)
+
+        return self._resolve_embedded_images(value, model)
 
     def _to_xml(self, value: markupsafe.Markup, /) -> str:
         return helpers.repair_html(value)
 
+    def _resolve_embedded_images(
+        self, value: str, model: capellambse.MelodyModel, /
+    ) -> markupsafe.Markup:
```
Wuestengecko, line 175: Same here, please use `self, model, value`.

Wuestengecko (commented):
Thanks for looking into this!

> An embedded image link contains the name of the Capella project as first element of the path in the `src` attribute

I think this first element here is the name of the "Eclipse-level" project, which is recorded in the `.project` file (and usually also used for the model itself, which is why `model.info.title` works - most of the time). This might be useful to figure out which resource to load from if there are multiple.

PR #466, opened 2024-11-01
feat(ci-templates): Use virtual environment

ewuerger:


Wuestengecko:
What's the need for this? Not necessarily against it, but I want to understand why / where we need this when it wasn't necessary before.

Wuestengecko:
Needed to support non-root Docker images / runners. Thanks for clarifying!

PR #474, opened 2024-11-14
Rename basic Accessors to make their roles clearer

Wuestengecko:
This commit performs the following renames:

- RoleTagAccessor to Containment
- AttrProxyAccessor to Association
- LinkAccessor to Allocation
- ReferenceSearchingAccessor to Backref

The old names of these Accessors will still be available at runtime until the end of v0.6.x.

### DIFF
--- a/capellambse/extensions/filtering.py
+++ b/capellambse/extensions/filtering.py
@@ -35,7 +35,7 @@ class FilteringCriterion(m.ModelElement):
 
     _xmltag = "ownedFilteringCriteria"
 
-    filtered_objects = m.ReferenceSearchingAccessor[m.ModelElement](
+    filtered_objects = m.Backref[m.ModelElement](
         (), "filtering_criteria", aslist=m.MixedElementList
     )
 
--- a/capellambse/extensions/pvmt/_config.py
+++ b/capellambse/extensions/pvmt/_config.py
@@ -265,18 +265,18 @@ class ManagedDomain(m.ModelElement):
     version = property(
         lambda self: self.property_values.by_name("version").value
     )
-    types = m.RoleTagAccessor(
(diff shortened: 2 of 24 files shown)


ewuerger:
If Containment is for the `RoleTagAccessor`, how do you plan to re-name `DirectProxyAccessor`? Essentially the elements accessed are defined under the element.
Also Association is only used when a relationship from an attribute is made. From `AttrProxyAccessor` this can be inferred.
I'd like to know the reasoning for this renaming.

Wuestengecko:
> If Containment is for the `RoleTagAccessor`, how do you plan to re-name `DirectProxyAccessor`?

It'll be dropped after #456 (most likely with v0.7). The implementation of DirectProxyAccessor was based on wrong assumptions, and now hinders development.

> Also Association is only used when a relationship from an attribute is made. From `AttrProxyAccessor` this can be inferred.

I'm not sure what you mean, do you mean the name "Association" is less clear?

> I'd like to know the reasoning for this renaming.

Make the relationship between these Accessor classes and the upstream metamodel attributes a bit clearer:

- `Containment` is used for `EReference` features with `containment="true"`.
- `Allocation` is used for ones that use `SomethingAllocation` helper objects, which conventionally are also called like `allocatedThings` in the upstream metamodel. (Note that there's usually also a Containment involved, which stores those helper objects - we currently don't expose those in the majority of cases.)
- `Association` is used for everything else, which use simple attributes with links to associate two (or more) objects.
- And finally, `Backref` is just much shorter and more convenient to type than `ReferenceSearchingAccessor`. :) As the other ones don't use the word "Accessor" anymore, it made sense to me to also rename this one at the same time.

ewuerger (approved):
LGTM.

PR #481, opened 2024-12-02
style: Fix HTML repr styling

huyenngn:
- Puts the icon and name of an element in the HTML repr title on one line
- Also makes title larger


Wuestengecko:
Before:

image

After:

image

> Puts the icon and name of an element in the HTML repr title on one line

But ... they were already in one line before.

> Also makes title larger

It's smaller now.

What exactly were you attempting to fix?

PR #494, opened 2024-12-18
Migrate to uv

Wuestengecko:
Use [uv](https://docs.astral.sh/uv/) for management of development dependencies and building in CI.

This also migrates the "docs" and "test" extras to dependency groups, as they were only used to facilitate the dev setup, and did not actually add any functionality.


ewuerger (changes requested):
I am very surprised how fast the environment setup with uv is. The installation of 77 packages in 49ms is insanely fast. The docs are almost instantly build. 

Can you explain why you modified the library structure by moving capellambse into a `src` folder? Is this necessary for uv to work? Or is this just best practice in order to exclude docs and tests from the build package?

I think you need to also check:
- CODEOWNERS (the paths aren't valid anymore)
- https://github.com/DSD-DBS/py-capellambse/blob/2e754e306fefd9b54f21eb1aa3bac6bc118f937f/src/capellambse/repl.py#L25
This doesn't work anymore, first the path needs to include ./src but also with `uv run` the tests aren't included anymore, so the test model can't be found. 

The rest looks fine and as soon as this is merged I'll start to migrate to uv in other projects.

Wuestengecko:
> Can you explain why you modified the library structure by moving capellambse into a `src` folder?

It's not strictly required for uv (in fact, uv doesn't care either way, it's up to the build backend to deal with this), but IMO this is a good opportunity for it. The two main reasons are:

1. It helps setuptools figure out what should actually go into the wheel and what not. We had problems with that in the past, where it unintentionally included megabytes worth of test models and docs stuff, and this is a good opportunity to fix that once and for all.

2. `uv init --lib` does it this way :)

> I think you need to also check:
> 
> * CODEOWNERS (the paths aren't valid anymore)
> 
> * https://github.com/DSD-DBS/py-capellambse/blob/2e754e306fefd9b54f21eb1aa3bac6bc118f937f/src/capellambse/repl.py#L25
>       
>   This doesn't work anymore, first the path needs to include ./src but also with `uv run` the tests aren't included anymore, so the test model can't be found.

Good catch! Fixed.

ewuerger (approved):
LGTM.

PR #500, opened 2025-01-14
docs: Add readthedocs config file

ewuerger:


docs/.readthedocs.yaml
Wuestengecko, line 1: We still have the `docs` job in GH actions as well, which should become redundant with this. Please remove that too.

Wuestengecko, line 1: Please also add the rest of the team to the RTD project, I think you're the only one with permissions there currently.

Wuestengecko, line 1: The RTD project also isn't hooked up to Github yet, so we don't see any status report in the PR. I want to at least see a green run before this can be merged.

Wuestengecko (changes requested):

docs/.readthedocs.yaml
ewuerger: I was only able to add you and @MoritzWeber0. @jamilraichouni and @vik378 didn't exist. Both the username or the e-mail.

ewuerger: The new host is: https://py-capellambse.readthedocs.io/en

There is also an option to set a custom and/or canonical URLs: https://docs.readthedocs.io/en/stable/custom-domains.html but I have no clue yet how this is done.

ewuerger: You see the new check on any new PR opened. [Here](https://github.com/DSD-DBS/py-capellambse/pull/502) for example. But it is red because this PR isn't merged yet.

Wuestengecko:
> Also the docs are now hosted on readthedocs.org.

On second thought, this does pose a problem in that every link / bookmark will break, without a good reason. Can we set RTD up to _only_ do PR builds, and keep GH pages for master? I think this would be a reasonable first step. That's also how it works over in https://github.com/DSD-DBS/capella-collab-manager.

ewuerger:
We can just keep the free additional hosting from rtd and still keep our docs hosted from github.

Wuestengecko (approved):

PR #504, opened 2025-01-20
feat: Align `Capability` and `CapabilityRealization`

ewuerger:
In order to enable `context_diagram` for `CapabilityRealization` in the end, we'll need the following exchanges:
- extends
- includes
- generalizes and their reverse

This PR aligns all Capability classes with each other, introducing a `CapabilityBase` class to stay dry.


Wuestengecko (changes requested):
Missing information about the metamodel should all be part of #477. If you need parts of it earlier, please backport from  the `extend-metamodel` branch instead. Or, even better, help get that code in a good enough shape to merge it :)

Wuestengecko:
Obsolete with merge of #477.

PR #508, opened 2025-02-12
Exit 3 on partial diagram cache export failure

Wuestengecko:
Signal to the caller that some diagrams failed to export by exiting with code 3. This allows automated scripts to more easily distinguish this case from other errors.

### DIFF
--- a/src/capellambse/diagram_cache.py
+++ b/src/capellambse/diagram_cache.py
@@ -100,6 +100,18 @@ else:
         with `capellambse.MelodyModel`.
 
         \b
+        Exit codes
+        ----------
+
+        The CLI will indicate the success status via exit codes:
+
+        \b
+        - 0 in case of success
+        - 1 for general errors
+        - 2 for CLI usage errors
+        - 3 if some diagrams failed to export, but others were successful
+
+        \b
         Refreshing representations
         --------------------------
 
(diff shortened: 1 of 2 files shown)


src/capellambse/diagram_cache.py
```diff
@@ -146,6 +146,8 @@ def _main(
             background=background,
             refresh=refresh,
         )
+        if any(not i["success"] for i in diagrams):
+            raise SystemExit(2)
```
jamilraichouni, line 150: Looks like exit code 2 is already reserved for a wrong usage of the CLI itself.

When I command

```shell
python -m capellambse.diagram_cache
```
I get the output

```
Usage: python -m capellambse.diagram_cache [OPTIONS]
Try 'python -m capellambse.diagram_cache --help' for help.

Error: Missing option '-m' / '--model'.
```
with an exit code of 2. But that would be misleading.

I found that there are exit codes with special meanings:

https://tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF

This is why we should use something else and I propose to go for exit code 3.

jamilraichouni (changes requested):

src/capellambse/diagram_cache.py
jamilraichouni: Other than that from above the `SystemExit` works. I checked it with a good real life example that led to this PR.
So when an exit code between three and 125 is chosen, I am happy with the changes in this PR.

jamilraichouni (approved):
LGTM

PR #511, opened 2025-02-19
Defer loading the filehandler until after `loadinfo`

Wuestengecko:
See also 0b4ca00d0cdc91c1267d4bd7aea3f5d20707ce16.

This keeps the returned modelinfo dict to simple basic types, which makes analyzing and logging easier downstream.


jamilraichouni (approved):
LGTM

PR #514, opened 2025-02-24
Treat more HTTP response codes as "not found"

Wuestengecko:
Some servers may send a different code than 404 if a file doesn't exist for some reason or another. For example, when requesting a PNG file from a diagram cache that only allows SVG, the cache server may respond with 400 Bad Request instead. Other servers may respond with 410 Gone if a file is requested that used to exist but has been deleted (because its retention period expired, for example).

### DIFF
--- a/src/capellambse/filehandler/http.py
+++ b/src/capellambse/filehandler/http.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import collections.abc as cabc
+import errno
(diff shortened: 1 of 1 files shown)


MoritzWeber0 (approved):
Works like a charm and makes it possible to use the the diagram cache API of the Collaboration Manager in Jupyter notebooks.

PR #518, opened 2025-03-17
Pre-resolve 'find', 'set' and 'extend' args during decl sync

Wuestengecko:
Before attempting to create a new object, try to pre-resolve all of the attributes that might need to be ready at time of creation.

This fixes an issue where objects could get created multiple times, each with only a subset of the requested operations applied to it.


vik378 (approved):
looks good!

PR #540, opened 2025-06-11
Reembed images in HTML PODs

### LINKED ISSUE #330: Handling of unembedded images
Capella 6.0+ stores embedded images external to the model XML, unlike previous version that used to have them embedded as `data:` URIs. We currently do not properly support this new storage scheme, which makes handling such descriptions with images annoying for library users.

Wuestengecko:
Since version 6.0, Capella prefers to place images as regular files into a directory next to the model. As the used paths are relative to the virtual Eclipse workspace root directory, this makes handling these images much more difficult.

With this commit, the text returned by HTMLStringPOD will have `` tags linking to a file replaced by ones containing the actual image data as base64-encoded data URI, with the original file path retained in the `data-capella-path` attribute. When writing to those attributes, these img tags are converted back, while writing modified image data either to the recorded file path or, if the `data-capella-path` attribute was removed, to a new file below `/images/` next to the model's entrypoint AIRD file.

---

A similar approach was explored earlier in #464. However, while that PR still required library users to explicitly deal with images in all cases, this PR allows to make image handling much more transparent.

Supersedes / Closes #464
Resolves #330

### DIFF
--- a/src/capellambse/helpers.py
+++ b/src/capellambse/helpers.py
@@ -14,6 +14,7 @@ import importlib.resources as imr
 import itertools
 import logging
 import math
+import mimetypes
 import operator
 import os
 import pathlib
@@ -25,6 +26,7 @@ import typing as t
 import urllib.parse
 import uuid
 
+import datauri
 import lxml.html
 import markupsafe
 import typing_extensions as te
@@ -34,6 +36,7 @@ from PIL import ImageFont
 
 import capellambse
 import capellambse._namespaces as _n
+import capellambse.filehandler as fh
 
 if sys.platform.startswith("win"):
     import msvcrt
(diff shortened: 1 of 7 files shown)


src/capellambse/helpers.py
```diff
@@ -578,6 +659,118 @@ def cb(el: etree._Element) -> None:
...
 
+def _get_handler(
+    resources: cabc.Mapping[str, fh.FileHandler], name: str, /
+) -> fh.FileHandler | None:
+    return (
+        resources.get(name)
+        or resources.get(name.removesuffix(".team"))
+        or resources.get("\x00")
+    )
+
```
ewuerger, line 670: I propose to add another helper function that reduces the duplication between both callback functions `cb` in embed and `unembed_images`:
```suggestion


def _get_handler_and_relpath(
    src: str, resources: cabc.Mapping[str, fh.FileHandler]
) -> tuple[fh.FileHandler, pathlib.PurePosixPath] | None:
    if not src or "://" in src:
        return None

    parts = pathlib.PurePosixPath(src).parts
    if len(parts) < 2:
        return None

    hdlname, *rest = parts
    relpath = pathlib.PurePosixPath(*rest)
    hdl = _get_handler(resources, hdlname)
    return (hdl, relpath) if hdl else None
```

src/capellambse/helpers.py
```diff
@@ -578,6 +659,118 @@ def cb(el: etree._Element) -> None:
...
+        try:
+            content = hdl.read_file(srcpath)
+        except FileNotFoundError:
+            return
+
+        b64content = base64.standard_b64encode(content).decode("ascii")
+        mime, _ = mimetypes.guess_type(srcpath, strict=False)
+        el.set("src", f"data:{mime};base64,{b64content}")
+        el.set("data-capella-path", src)
```
ewuerger, line 711:
```suggestion
    def cb(el: etree._Element) -> None:
        if el.tag != "img":
            return
        src = el.get("src") or ""
        resolved = _get_handler_and_relpath(src, resources)
        if not resolved:
            return
        hdl, relpath = resolved
        try:
            content = hdl.read_file(relpath)
        except FileNotFoundError:
            LOGGER.debug("Image not found: %s", relpath)
            return
        mime, _ = mimetypes.guess_type(relpath.as_posix(), strict=False)
        data_uri = _bytes_to_data_uri(
            content, mime or "application/octet-stream"
        )
        el.set("src", data_uri)
        el.set("data-capella-path", src)
```

src/capellambse/helpers.py
```diff
@@ -578,6 +659,118 @@ def cb(el: etree._Element) -> None:
...
+            ptree = etree.fromstring(dotproject)
+            try:
+                (hdlname,) = ptree.xpath("/projectDescription/name/text()")
+            except ValueError:
+                return
+
+        hdl.write_file(srcpath, src.data)
+        el.set("src", f"{hdlname}/{srcpath}")
+        with contextlib.suppress(KeyError):
+            del el.attrib["data-capella-path"]
```
ewuerger, line 769:
```suggestion
    def cb(el: etree._Element) -> None:
        if el.tag != "img":
            return

        src = el.get("src") or ""
        if not src.startswith("data:"):
            return

        data = datauri.DataURI(src)
        if capella_path := el.get("data-capella-path"):
            resolved = _get_handler_and_relpath(capella_path, resources)
            if not resolved:
                return
            hdl, relpath = resolved
            hdlname = pathlib.PurePosixPath(capella_path).parts[0]
        else:
            ext = (
                mimetypes.guess_extension(data.mimetype or "", strict=False)
                or ""
            )
            relpath = pathlib.PurePosixPath("images") / f"{generate_id()}{ext}"
            hdl = resources.get("\x00")
            if not hdl:
                return
            try:
                dot = hdl.read_file(".project")
                tree = etree.fromstring(dot)
                hdlname = tree.xpath("/projectDescription/name/text()")[0]
            except Exception:
                LOGGER.debug("Could not read project name for default handler")
                return

        hdl.write_file(relpath, data.data)
        el.set("src", f"{hdlname}/{relpath.as_posix()}")
        el.attrib.pop("data-capella-path", None)
```

src/capellambse/helpers.py
ewuerger: The `_bytes_to_data_uri` function` which I find actually useful and maybe want to reuse it in other projects too. So maybe it shouldn't have the leading `_`:
```
def _bytes_to_data_uri(data: bytes, mime: str) -> str:
    """Encode raw bytes as a base64 data URI with given MIME type."""
    b64 = base64.standard_b64encode(data).decode("ascii")
    return f"data:{mime};base64,{b64}"
```

ewuerger (changes requested):
With the proposed changes we remove unnecessary duplication between the embed and unembed callback functions like parsing the `src`, finding the handler, read/write bytes and guessing the mimetypes. Also `FileNotFoundError`s should be logged atleast at debug level. Maybe another level is more appropriate (ERROR)? Manual parts slicing on `pp` (cool name btw xD) was simplified too. And optionally `_bytes_to_data_uri` defines a nice utility function for converting bytest to a DataURI string/path. 

Of course if you don't like the `_bytes_to_data_uri` then leave it out. After resolving these issues I'm fine and LGTM.

ewuerger (approved):
LGTM.

PR #548, opened 2025-06-20
Run tests in parallel with pytest-xdist

Wuestengecko:
This obviously makes running tests much faster.

To avoid problems with local IDE-/editor-integrated test runners, parallelization is only active by default in CI. However, local runs can easily be parallelized using `pytest -n auto` instead of plain `pytest`.

CI times are cut almost exactly in half, from ~70 minutes cumulative on current master to ~36 minutes.


ewuerger (approved):
LGTM.

PR #549, opened 2025-06-23
fix: Support multiple enums under scopes

ewuerger:
With this fix multiple class-rules under the scope for different classes are supported.


Wuestengecko:
Looks good at first glance, can you please add the `(pvmt)` scope to the commit description?

PR #556, opened 2025-07-01
Improve documentation of `MelodyModel.update_diagram_cache()`

Wuestengecko:
Add more concrete example calls and update the existing ones so they will actually work.


ewuerger (approved):
LGTM.

PR #557, opened 2025-07-01
feat(ci-templates): Avoid multi-line script blocks in Gitlab templates

Wuestengecko:
Gitlab only shows the first line of a script block in the CI output if it spans multiple lines. This makes it more difficult to understand what the CI run was trying to do, especially if multiple things happened inside the script block, so avoid using multi-line blocks where possible. Where it's not possible, add some log statements to make debugging at least a little bit easier.


ewuerger (approved):
LGTM.

PR #619, opened 2025-12-22
feat: added support for the libraries linking and referencing from model

wingedfox:
Added method into the loader to support libraries linking.
Use of the resources and referencing them in the extensions does not work. lookup fails with missing fragment.
Here I fix this by allowing explicit library loading and linking into a model.


src/capellambse/model/_descriptors.py
```diff
@@ -363,6 +364,90 @@ def __repr__(self) -> str:
...
+        """Retrieve the value of the attribute."""
+        if obj is None:
+            return self
+
+        objs: t.Any = None
+        try:
+            objs = self.wrapped.__get__(obj, type(obj))
+        finally:
+            self.__objs = objs
+            return objs
```
Wuestengecko: - The accessor instance lives as long as the Python process, which means we can't use it to store information about a specific model.
- We can't rely on `__get__` being called before `__set__` or `__delete__` anyway.
- This `try / finally` effectively works like `try / except BaseException`, which would make the intent clearer, and would also allow narrowing down to at least `Exception` (instead of `BaseException`, which includes things like `KeyboardInterrupt` and `SystemExit`).

src/capellambse/model/_descriptors.py
```diff
@@ -363,6 +364,90 @@ def __repr__(self) -> str:
...
+    """
+
+    def __init__(
+        self,
+        wrapped: Accessor[T_co | _obj.ElementList[T_co] | None ],
+        optional: bool = False
+    ) -> None:
+        """Create a new optional descriptor."""
+        self.wrapped: t.Final = wrapped
+        self.optional: t.Final = optional
```
Wuestengecko: `self.optional` seems like a development left-over. (It's also mentioned in the docstring.)

Wuestengecko (changes requested):
I'm a bit confused on who this API is for and when/how it should be used:

- Is it for linking in an entirely new library? In the docstring example, `project.extensions` is already pre-filled, which is confusing in that case.
- Is it for repairing a broken model? Then it'd be good to have some documentation on the symptoms of brokenness (also how it shows up in Capella, not just py-capellambse), and possibly create a standalone script that can directly be called (as I mentioned in a comment).
- Is it meant to be called by some internal APIs/functions that don't exist yet / haven't been adjusted to call it yet?

wingedfox:
> * .

Yes, it targets use of entirely new libraries without running Capella to link library and operate REC/RPL in the model.
Adding library to the model resources does not actually update model fore the future library reference. Neither it adds new fragments to the model tree.

This PR does not target model fixing, but adds new function of
* Adding libraries (referencing in the resources)
* Fixing REC Catalog operations (Filter Extension crashes on model extension access)
* Fixing Links between the model elements (changes to the Single on attributes)

It seems nobody needed this before that's why relatively simple PR introduced many changes.

Basic scenario is
1. Add new Library (referenced in the model resources) into model
2. Create relevant REC Catalog
3. Navigate into Library
4. Copy library elements into target model
5. Setup links between the model elements

src/capellambse/loader/core.py
```diff
@@ -590,6 +591,85 @@ def check_duplicate_uuids(self) -> None:
...
+            If metadata could not be found
+        """
+        metadata = next(afm.root.iter(METADATA_TAG), None)
+        if metadata is None:
+            raise RuntimeError("Cannot find <Metadata> in primary .afm file")
+        LOGGER.debug("Found <Metadata> with ID %s", metadata.get("id"))
+        return metadata
+
+    def _link_library(
+        self, lib: pathlib.PurePosixPath
```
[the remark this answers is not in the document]
wingedfox: It's not obvious at all, for instance, in my setup the structure is the following
* libs
  * lib1
  * lib2
* proj1 -> resources: ../libs/lib1
* proj2 -> resources: ../libs/lib2
what is a proper way to handle this?

src/capellambse/extensions/filtering.py
```diff
@@ -142,8 +142,10 @@ class IntersectionFilteringResultSet(FilteringResultSet):
 
 
 def init() -> None:
-    capellamodeller.SystemEngineering.filtering_model = m.Single(
-        m.Filter("extensions", (NS, "FilteringModel"))
+    capellamodeller.SystemEngineering.filtering_model = m.Optional(
+        m.Single(
+            m.Filter("extensions", (NS, "FilteringModel"))
+        )
     )
```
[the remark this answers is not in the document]
wingedfox: Yes, it's the only place at the moment. The problem behind is that FilteringModel without Optional crashes on touch of the extensions - `model.project.extensions` if filtering extension was not enabled.
This effectively blocks Libraries, because they are the extensions as well, i.e.
```
  <ownedExtensions xsi:type="libraries:ModelInformation" id="a...a">
    <ownedReferences xsi:type="libraries:LibraryReference" id="9...d"
        library="libraries:ModelInformation ../lb1/lib1.capella#b...e"/>
  </ownedExtensions>
```

src/capellambse/loader/core.py
```diff
@@ -590,6 +591,85 @@ def check_duplicate_uuids(self) -> None:
...
+
+        lib = model.project.extensions[0].reference.library #no longer crashes, all fragments are in place
+
+        ```
+        """
+        handler = self.resources[str(lib)]
+        _h , filename = _derive_entrypoint(handler)
+        frag = ModelFile(
+            filename, handler, ignore_uuid_dups=self.__ignore_uuid_dups
+        )
```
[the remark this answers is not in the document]
wingedfox: ok, got it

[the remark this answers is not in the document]
wingedfox: fixed

src/capellambse/metamodel/capellamodeller.py
```diff
@@ -57,6 +58,10 @@ class SystemEngineering(capellacore.AbstractModellingStructure, ModelRoot):
     [source:MIL-STD 499B standard]
     """
 
+    extensions = m._descriptors.Containment[re.ReElementContainer](
+        "ownedExtensions", (ns.RE, "RecCatalog")
+    )
```
Wuestengecko: `extensions` is already defined on `ModelElement` to take any type of element (because we might not even know the actual type yet), and to my understand of the metamodel, this is the best thing we can do. Limiting it here to only accept objects of type `RecCatalog` would mean that adding any extension type other than that would be illegal and render the model corrupt.

If you want to filter the extensions for convenient access to only catalogs, use `m.Filter` on the existing `extensions` attribute instead:

    rec_catalogs = m.Filter[re.ReElementContainer]("extensions", (ns.RE, "RecCatalog"))

src/capellambse/loader/core.py
```diff
@@ -590,6 +591,85 @@ def check_duplicate_uuids(self) -> None:
...
+                meta_lib = self.__get_metadata(self.trees[ref_name])
+                meta_self = self.__find_metadata()
+
+                if not next(filter(lambda el: re.search(str(ref_name), el.attrib["href"]), meta_self.iterchildren("additionalResources")), None):
+                    ael = meta_self.makeelement("additionalResources", href=f"../{ref_name}#{meta_lib.attrib['id']}")
+                    meta_self.append(ael)
+            elif ref_name.suffix == ".capella":
+                aird_self = self.trees[pathlib.PurePosixPath(f"\x00/{self.entrypoint}")]
+
+                last = next(filter(lambda el: re.search(str(ref_name), el.text), XP_SEMANTIC_RESOURCES(aird_self.root)), None)
```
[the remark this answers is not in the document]
wingedfox: it's doesn't matter, at this point we have to find the last reference
of course it would be better to operate Capella DOM, but it's different and more complex story

src/capellambse/loader/core.py
```diff
@@ -32,6 +32,7 @@
 
 import capellambse._namespaces as _n
 from capellambse import filehandler, helpers
+from capellambse.aird._common import XP_SEMANTIC_RESOURCES
```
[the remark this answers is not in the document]
wingedfox: fixed

src/capellambse/metamodel/capellamodeller.py
wingedfox: removed overload, but need to think about more convenient filtering in the derived products

wingedfox:
Here's the reproduced bug related to broken Filtering Extension and why Optional was needed to hide this error
```
<class 'capellambse.metamodel.capellamodeller.SystemEngineering'>

✗ Error: Required viewpoint not activated: None
Traceback (most recent call last):
  File "/../cli.py", line 30, in cli
    print (x)
    ~~~~~~^^^
  File "/../site-packages/capellambse/model/_obj.py", line 810, in __repr__
    value = getattr(self, attr)
  File "/../site-packages/capellambse/model/_descriptors.py", line 422, in __get__
    objs: t.Any = self.wrapped.__get__(obj, type(obj))
                  ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/../site-packages/capellambse/model/_descriptors.py", line 2659, in __get__
    cls = obj._model.resolve_class(self.class_)
  File "/../site-packages/capellambse/model/_model.py", line 772, in resolve_class
    raise RuntimeError(
        f"Required viewpoint not activated: {viewpoint!r}"
    )
RuntimeError: Required viewpoint not activated: None
```

Wuestengecko:
Looks like the last merge has gone wrong a bit and a lot of commits are now duplicated. Could you please run these commands locally, which should fix it? (Make sure to commit or stash first, so that `git status` shows no changes)

````sh
git switch -d ba9639869
git pull --no-ff https://github.com/dbinfrago/py-capellambse master
git checkout --theirs src/capellambse/metamodel/pa/__init__.py
git add src/capellambse/metamodel/pa/__init__.py
git merge --continue
git rebase 29b9cee8f master --onto @
git switch master
````

File: src/capellambse/loader/core.py [hunk 9]
```diff
@@ -590,6 +591,85 @@ def check_duplicate_uuids(self) -> None:
...
+        """
+        metadata = next(afm.root.iter(METADATA_TAG), None)
+        if metadata is None:
+            raise RuntimeError("Cannot find <Metadata> in primary .afm file")
+        LOGGER.debug("Found <Metadata> with ID %s", metadata.get("id"))
+        return metadata
+
+    def _link_library(
+        self, lib: pathlib.PurePosixPath
+    ) -> None:
```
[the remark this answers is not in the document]
Wuestengecko: Just remove the leading underscore to mark it public, i.e. `def link_library`. Then it should also show up in the Sphinx docs automatically :)

src/capellambse/metamodel/information/datavalue.py
```diff
@@ -236,11 +236,11 @@ class AbstractExpressionValue(
...
     operator = m.EnumPOD("operator", BinaryOperator)
-    left_operand = m.Containment["DataValue"](
-        "ownedLeftOperand", (NS, "DataValue")
+    left_operand = m.Single["DataValue"](
+        m.Containment["DataValue"]("ownedLeftOperand", (NS, "DataValue"))
     )
-    right_operand = m.Containment["DataValue"](
-        "ownedRightOperand", (NS, "DataValue")
+    right_operand = m.Single["DataValue"](
+        m.Containment["DataValue"]("ownedRightOperand", (NS, "DataValue"))
```
Wuestengecko: We only need type annotations on the outer `Single`, the inner `Containment` (or `Association`/...) can then be inferred from that.

<sub>(Be careful when clicking "commit" here, I think the suggestions window is broken again.)</sub>

```suggestion
    left_operand = m.Single["DataValue"](
        m.Containment("ownedLeftOperand", (NS, "DataValue"))
    )
    right_operand = m.Single["DataValue"](
        m.Containment("ownedRightOperand", (NS, "DataValue"))
```

src/capellambse/metamodel/re.py
```diff
@@ -55,12 +55,16 @@ class GroupingElementPkg(CatalogElementPkg):
...
 
 class CatalogElementLink(ReAbstractElement):
-    source = m.Association["CatalogElement"]((NS, "CatalogElement"), "source")
-    target = m.Association["m.ModelElement"](
-        (ns.MODELLINGCORE, "ModelElement"), "target"
+    source = m.Single(m.Association["CatalogElement"]((NS, "CatalogElement"), "source"))
+    target = m.Single(
+        m.Association["m.ModelElement"](
+            (ns.MODELLINGCORE, "ModelElement"), "target"
+        )
```
Wuestengecko: When wrapping new attributes with `Single`, please also move the annotations to the outermost descriptor. This often makes it look a little nicer if things don't quite fit on one line :)


```suggestion
    source = m.Single["CatalogElement"](m.Association((NS, "CatalogElement"), "source"))
    target = m.Single["m.ModelElement"](
        m.Association((ns.MODELLINGCORE, "ModelElement"), "target")
```

Wuestengecko (commented):
I'll also do some more thorough testing in the next days, but in general it looks mostly good - is the PR ready from your perspective, or do you have more to implement?

src/capellambse/loader/core.py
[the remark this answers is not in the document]
wingedfox: fixed

wingedfox:
I've been working on the very deep model modifications and I'm not sure what I can find and fix.

For instance, I've found that in some cases asking collection to add IntegerPropertyValue I get String instead, in other cases - Boolean. 

I miss methods to "copy" elements and element creation for non-collection elements like ownedMinCard/ownedMaxCard.
Found that wrap_xml doesn't dare to check copied properties against model and Capella crashes due to unresolved references.

Hope we can merge this PR before anything else come, I have reached 95% MVP readiness with current PR.

What I've been missing a lot is the new ELK routing, because my models have something around 11K elements per model with around 20 Components/100 functions/500 exchanges per layer.

wingedfox:
here's the article and tool based on py-capellambse
https://www.linkedin.com/pulse/federative-mbse-system-architecture-ilya-lebedev-op3af
https://github.com/open-modeling/moirai-se

wingedfox:
I've been waiting for a resolution of this PR...

wingedfox:
@Wuestengecko @ewuerger could you please take a look?

wingedfox:
@Wuestengecko please check the pipiline

wingedfox:
@Wuestengecko please check the PR and pipeline

wingedfox:
@Wuestengecko please check PR

src/capellambse/loader/core.py
```diff
@@ -1302,6 +1407,15 @@ def _find_fragment(
...
                 return (fragment, tree)
+        # fallback here is necessary because
+        # working with root collections, i.e. key_value_pairs,
+        # causes project root being represented by two entities, i.e.
+        # <Element {http://www.polarsys.org/capella/core/modeller/7.0.0}Project at 0x7feced455dc0>
+        # <Element {http://www.polarsys.org/capella/core/modeller/7.0.0}Project at 0x7fecee0f6640>
+        # they do not pass "is" check but have the same IDs
+        for fragment, tree in self.trees.items():
+            if tree.root.attrib.get("id") == root.attrib.get("id"):
+                return (fragment, tree)
```
Wuestengecko, line 1418: This sounds odd. There should only ever be exactly one LXML entity per model element, _especially_ the project root (under which everything else is nested).

- Does this refer to the root of the (newly linked) library project? In this case, we should find it in `self.trees` - if not, we're missing the entry here.

- If it actually is the same model, we have accidentally duplicated the project tree somehow (though it's not obvious to me how that would happen right now), and are now mixing elements from both tree instances.

- Could this be due to some buggy interaction (or perhaps race) with the namespace map update code? (Be aware that capellambse currently isn't thread-safe, so this might be caused by multi-threaded downstream code.)

All of these could easily lead to corrupted models and data loss, so this needs to be addressed before merging. If you can, could you add a test case which hits this case?

src/capellambse/loader/core.py
```diff
@@ -590,6 +595,106 @@ def check_duplicate_uuids(self) -> None:
...
+        When you need to refere to external library (or reuse one in a project)
+        ```
+        p = "..." #path to library
+        model._loader.link_library(p)
+
+        lib = model.project.extensions[0].reference.library
+        ```
+        """
+        handler = self.resources[str(lib)]
+        _h, filename = _derive_entrypoint(handler)
```
Wuestengecko, line 643:
```suggestion
        _, filename = _derive_entrypoint(handler)
```

src/capellambse/loader/core.py
```diff
@@ -590,6 +595,106 @@ def check_duplicate_uuids(self) -> None:
...
+        Method updates .aird, .capella, .afm to correcrly reflect library in target model
+
+        When you need to refere to external library (or reuse one in a project)
+        ```
+        p = "..." #path to library
+        model._loader.link_library(p)
+
+        lib = model.project.extensions[0].reference.library
+        ```
+        """
```
Wuestengecko, line 641: The docs build is failing because sphinx doesn't like this docstring in particular.

- There's no "Description" heading. The full description goes right after the one-line summary, separated by a single blank line and no extra headings in between.

- Can you move the code example into an "Examples" section after "Parameters"? I think this is where it'd fit best in this case.

- Could you please also add a test case (or multiple) which exercises this new code? This would also serve nicely as secondary documentation. We already have a few test models in the `conftest.Models` dict which should come in handy here.

You can test the docs build locally as well (which should be much faster than waiting for CI) with a simple `make docs`.

src/capellambse/metamodel/capellamodeller.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import capellambse.model as m
+from capellambse.metamodel import re
```
Wuestengecko, line 6: I don't know why Github isn't running CI for this PR. Could you please make sure you run all the pre-commit hooks locally with `pre-commit run -a` ? Here we have an unused import which the hooks would warn about (and automatically remove for you).

src/capellambse/metamodel/interaction.py
```diff
@@ -83,9 +83,11 @@ class SequenceMessage(capellacore.NamedElement):
...
     )
-    sending_end = m.Association["MessageEnd"]((NS, "MessageEnd"), "sendingEnd")
-    receiving_end = m.Association["MessageEnd"](
-        (NS, "MessageEnd"), "receivingEnd"
+    sending_end = m.Single(
+        m.Association["MessageEnd"]((NS, "MessageEnd"), "sendingEnd")
+    )
+    receiving_end = m.Single(
+        m.Association["MessageEnd"]((NS, "MessageEnd"), "receivingEnd")
     )
```
Wuestengecko, line 91: Please move the bracketed type annotation up a level to the `m.Single` instance instead of whatever it wraps. This helps keep things nice and consistent (i.e. it's always the outermost descriptor that has the annotation).

```diff
-m.Single(
-    m.Association["Annotation"](...)
+m.Single["Annotation"](
+    m.Assocation(...)
 )
```

Same also in the other metamodel files you modified.

src/capellambse/loader/core.py
```diff
@@ -590,6 +595,106 @@ def check_duplicate_uuids(self) -> None:
...
+                meta_lib = self.__get_metadata(self.trees[ref_name])
+                meta_self = self.__find_metadata()
+
+                if not next(
+                    filter(
+                        lambda el: re.search(str(ref_name), el.attrib["href"]),
+                        meta_self.iterchildren("additionalResources"),
+                    ),
+                    None,
+                ):
```
Wuestengecko, line 671: We're doing a simple substring match here, no need to use a regex. (Same applies a few lines further down for `.capella` files.)

Also, the `next(filter(lambda el: ..., ...), None)` construct can be rewritten using `any()` and a generator expression, which saves a little bit of overhead.

```suggestion
                if not any(
                    str(ref_name) in el.attrib["href"]
                    for el in meta_self.iterchildren("additionalResources")
                ):
```

src/capellambse/loader/core.py
```diff
@@ -590,6 +595,106 @@ def check_duplicate_uuids(self) -> None:
...
+                    pathlib.PurePosixPath(f"\x00/{self.entrypoint}")
+                ]
+
+                last = next(
+                    filter(
+                        lambda el: re.search(str(ref_name), el.text),
+                        XP_SEMANTIC_RESOURCES(aird_self.root),
+                    ),
+                    None,
+                )
```
Wuestengecko, line 688: The same points as above also apply here (regex for substring match and filter+lambda overhead). Also, here we're doing the XPath lookup twice, which I think we could combine into a single iteration like this (notice that the `else` belongs to the loop itself, not the enclosed `if`):

```python
for el in XP_...:
    if str(ref_name) in el.text:
        break
else:
    sr = el.makeelement(...) # etc.
```

src/capellambse/model/_pods.py
```diff
@@ -495,7 +495,6 @@ def __setitem__(
             it = self._parent.iterchildren(self._tag)
             elems = list(itertools.islice(it, idx.start, idx.stop, idx.step))
             values = [i.text for i in elems]
-            values[idx] = value
             for e, v in zip(elems, values, strict=True):
                 e.text = v
```
Wuestengecko, line 500: Good catch, this is/was indeed bugged. But I think the fix is incomplete: Now the `for` loop is just setting the current values instead of the new ones. Instead, we can simply iterate over the `islice` directly, i.e.:

```python
elems = itertools.islice(it, idx.start, idx.stop, idx.step)
for e, v in zip(elems, value, strict=True):
    e.text = v
```

src/capellambse/loader/core.py
```diff
@@ -590,6 +595,106 @@ def check_duplicate_uuids(self) -> None:
...
+            metadata element if found
+
+        Raises
+        ------
+        RuntimeError
+            If metadata could not be found
+        """
+        metadata = next(afm.root.iter(METADATA_TAG), None)
+        if metadata is None:
+            raise RuntimeError("Cannot find <Metadata> in primary .afm file")
```
Wuestengecko, line 618: This exception message doesn't quite fit anymore, since this function is also used for auxiliary `.afm` files. Let's simply show the filename instead:


```suggestion
        if metadata is None:
            raise RuntimeError("Cannot find <Metadata> in {afm.filename}")
```

Wuestengecko (commented):
Hi! Sorry for coming back to you so late.

I do have a few more comments. Especially the duplicated tree part looks concerning, we should definitely take a closer look at what's causing that. I think it'd be great to have a few additional test cases there (and for the library code more generally; we don't really have a lot of tests for that part at all yet).

PR #632, opened 2026-01-20
Drop Python 3.10

Wuestengecko:
This is mainly done for the sake of uv, which requires that all dependencies (even development-only ones) have to be resolvable for all allowed Python versions, even if those Python versions are not actually used. After capellambse-context-diagrams has dropped 3.10 support a while ago already, keeping support in capellambse doesn't make much sense anymore either.


ewuerger (approved):
LGTM.

PR #656, opened 2026-08-26
feat: Add functional chain icon

ewuerger:


pyproject.toml
```diff
@@ -155,6 +155,7 @@ typecheck = [
   "jinja2==3.1.6",
   "markupsafe==3.0.3",
   "mypypp==0.1.1",
+  "numpy<2.5",
```
Wuestengecko: Image

ewuerger: Yes. Mypy complained. I had to.

Wuestengecko: > .venv/lib/python3.12/site-packages/numpy/\_\_init\_\_.pyi:737: error: Type statement is only supported in Python 3.12 and greater  [syntax]

Let's not go overboard and limit everyone downstream just because we still want to support some older Python version. (To be fair, it should be fine to drop 3.11 about now anyway, but that's something for a different PR.)

```suggestion
  "numpy<2.5 ; python_full_version < '3.12'",
```

